PR #8, opened 2016-06-16
fixed zeroconf compile errors

penfold42:
zeroconf wouldn't compile.


redPanther:
just a note - this is already fixed in beta

penfold42:
Should we consider ditching the beta branch ?

Could we not use tags or release branches and use master as the latest ?

redPanther:
I would appreciate that

brindosch:
yay of course

Issue #4, opened 2016-06-16
UDP


brindosch:
- [x] refactor UDP as seperate listener and remove it from the bootsequence and effect list

UDP should be a listener (clean) implemented like boblight
```
	"udpServer" : 
	{
		"enable" : false,		
		"port" : XXXXX,
		"priority" : 900
	},
```

penfold42:
I'll do a native listener

penfold42:
I've started.... https://github.com/penfold42/hyperion.ng/tree/udp_listener

I've literally cloned the boblight server and done one hell of a search and replace.
In fact, here's roughly what i did in case its helpful for someone else...

cp -a boblightserver/ udplistener/
mv BoblightClientConnection.cpp	UDPClientConnection.cpp
mv BoblightServer.cpp	UDPListener.cpp

mv BoblightClientConnection.h	UDPClientConnection.h
mv BoblightServer.h UDPListener.h


%s/BoblightServer/UDPListener/g
%s/BoblightClientConnection/UDPClientConnection/g
%s/boblightserver/udplistener/g
%s/BoblightClientConnection.cpp/UDPClientConnection.cpp/g
%s/BoblightClientConnection.h/UDPClientConnection.h/g
%s/BoblightServer/UDPListener/g
%s/boblightServer/udpListener/g
%s/boblightserver/udplistener/g
%s/BOBLIGHT/UDPLISTENER/g

brindosch:
Great! If it works just remove the code from the effectengine (?) :)

redPanther:
there wasn't any code regarding udp in effect engine. just a additional python script

penfold42:
Correct - a hastily cobbled together Python script that I thought was worth sharing (and deleting shortly!)

penfold42:
#48 
I think we can close this issue now?

brindosch:
y

PR #18, opened 2016-06-20
New builtin udp listener

penfold42:
(this replaces the other pull request that for some unknown reason wasnt clean)

1. Tell us something about your changes.
Add a native builtin UDP listener to address #4 

2. If this changes affect the .conf file. Please provide the changed section
"udpListener" :
{
"port" : 2801,
"priority" : 890,
"timeout" : 5000
},

port is the udp port to listen on
priority - defaults to 890
timeout - if we see not packet for 5000 ms, clear the priority


redPanther:
Thx for implenting! Did you plan to integrate other udp protcols like tpm2net in the listener? I think this vould be a cool feature. I think the windows software glediator can write tpm2net nativly

penfold42:
its not at the top of my list!

brindosch:
Great! i have some question.

- The udp listener starts always and you can´t disable it - you need to wait after startup 5secs until you could use another component for example with prio 900? Add enable disable option?
- We should talk about a useful prio order in general, our current options should not collide with each other by default.
- i like the idea of timeout, should be also implemented for boblight? cause this component will also block the possibilities for a "idle effect" (if no packages received) (redPanther may hate me for that description ;) )

redPanther:
Blocking things schould be implemented in threads

penfold42:
ill add the enable/disable - can you merge whats there ?

with timezone differences we are more likely to trip over each other - I had to rework it already with the refactoring (which is good!)

brindosch:
Yes, you are at the wrong time zone ;)

PR #22, opened 2016-06-20
add "enable" : true/false to both udp listener and boblight, zeroconf cleanup

penfold42:
added "enable" : true/false to both udp listener and boblight
Changed the zeroconf/mDNS service names

        "boblightServer" :
        {
                "enable" : false,
                "port" : 19333
        },

        "udpListener" :
        {
                "enable" : true,
                "port" : 2801,
                "priority" : 880,
                "timeout"  : 5000
        },

You can no longer configure the mDNSDescr nor mDNSname from the json configuration

### DIFF
--- a/src/hyperiond/hyperiond.cpp
+++ b/src/hyperiond/hyperiond.cpp
@@ -218,26 +218,52 @@ void HyperionDaemon::startNetworkServices()
 	if (_config.isMember("boblightServer"))
 	{
 		const Json::Value & boblightServerConfig = _config["boblightServer"];
-		_boblightServer = new BoblightServer(hyperion, boblightServerConfig.get("priority",900).asInt(), boblightServerConfig["port"].asUInt());
-		Info(_log, "Boblight server created and started on port %d", _boblightServer->getPort());
+		if ( boblightServerConfig.get("enable", true).asBool() )
+		{
+			_boblightServer = new BoblightServer(hyperion,
+						boblightServerConfig.get("priority",900).asInt(),
+						boblightServerConfig["port"].asUInt()
+						);
+			Info(_log, "Boblight server created and started on port %d", _boblightServer->getPort());
+		}
 	}
 
         // Create UDP listener if configuration is present
         if (_config.isMember("udpListener"))
         {
                 const Json::Value & udpListenerConfig = _config["udpListener"];
-                _udpListener = new UDPListener(hyperion,
-                                        udpListenerConfig.get("priority",890).asInt(),
-                                        udpListenerConfig.get("timeout",10000).asInt(),
-                                        udpListenerConfig["port"].asUInt()
-                                );
-                Info(_log, "UDP listener created and started on port %d", _udpListener->getPort());
+		if ( udpListenerConfig.get("enable", true).asBool() )
+		{
+			_udpListener = new UDPListener(hyperion,
+						udpListenerConfig.get("priority",890).asInt(),
+						udpListenerConfig.get("timeout",10000).asInt(),
+						udpListenerConfig.get("port", 2801).asUInt()
+					);
+			Info(_log, "UDP listener created and started on port %d", _udpListener->getPort());
+		}
         }
 
 	// zeroconf
 	const Json::Value & deviceConfig = _config["device"];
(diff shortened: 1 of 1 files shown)


brindosch:
wow, fast and with zeroconf, great!
you should set the same name pattern for all 3 zeroconf entrys. Maybe you could add your final decision to all 3 :)
I don´t know what is best practice here. maybe @redPanther could say something.

hyperiond_json hyperiond_proto hyperiond_udp

Do we need that much ___. I think about the discovery i have no idea how a app perform this? a matching string? or half invalid and need to do regex to transform this? No idea :)

redPanther:
I would suggest:
description: deviceName@host
service: _hyperiond-service._protocol

e.g. my printer make this:
Brother DCP-L2540DN series                    _scanner._tcp
or my synology nas:
Lager                                         _sftp._tcp

penfold42:
I just saw lots of other services doing the _ thing !

run avahi-browse -a to look at all the services on your network.

Run avahi-browse -r _hyperiond_json._tcp to find that specific service on the network

penfold42:
Agreed I only just noticed that xbmc uses a hyphen rather be than an underscore

should we strip out the configuration options ?
Maybe wait a month or until someone writes a client that cares about them ?

redPanther:
I would kill the config options. I'm not shure, but I think it's not important if we use hyphen or underscore.

I would change zeroconf code to this:


```
	// zeroconf
	const Json::Value & deviceConfig = _config["device"];
	const std::string deviceName = deviceConfig.get("name", "").asString();
	const std::string mDNSDescr = deviceName+"@"+QHostInfo::localHostName().toStdString();

	// zeroconf json
	BonjourServiceRegister *bonjourRegister_json = new BonjourServiceRegister();
	bonjourRegister_json->registerService(BonjourRecord(mDNSDescr.c_str(), "_hyperiond_json._tcp",QString()), _jsonServer->getPort() );
	Info(_log, "Json mDNS responder started");

	// zeroconf proto
	BonjourServiceRegister *bonjourRegister_proto = new BonjourServiceRegister();
	bonjourRegister_proto->registerService(BonjourRecord(mDNSDescr.c_str(), "_hyperiond_proto._tcp",QString()), _protoServer->getPort() );
	Info(_log, "Proto mDNS responder started");

```
clean and simple ;-)

BTW I see you using Hyperion* as argument. Would be good if you can remove that and use Hyperion::getInstance() in your class (you can use it directly in your initializer if you want: _hyperion(Hyperion::getInstance()) )

Boblight using still the hyperion argument, but this should be removed too.

redPanther:
I log level "info" good here? I can imagine that "Debug" fits more in that situation

brindosch:
> should we strip out the configuration options ?

I would say yes, cause what could you reach with that option? A not detected Hyperion is the only -real-one?
We should set this as a "always the same string". I am sure kodi never changed this since 10 years or so :) every app would brake if people don´t follow our "rules" and change it.
no option - no error. Keep it simple :)

penfold42:
im removing the options right now.... hold off on the merge

(i put the there coz i had no idea what they should be!)

penfold42:
good to merge i think

@redPanther I'll change the Hyperion* a bit later - feel free if you're motivated.
I blame you..... you should have refactored boblight before I based this on it! *grin*

brindosch:
you changed back the prio of udplistener to 890. Was not your intention?

penfold42:
I cant remember !

It has to be less than the grabber, or it will never win.
890 is the default right now

brindosch:
collides with framegrabber which we use currently 890

penfold42:
We should probably create a default_priorities.h

800 ?

brindosch:
yeah the priority juggle is funny and you have no real overview.
sounds good :)
grabbers around 900, network data/proto around 800, booteffect is at 700 (to see it really, when you enable it). yeah its okay i think :)

penfold42:
external grabbers use 800 as the default.

I'll use 700 if no-one objects in the next 2 minutes!

brindosch:
ah right, but we remove the interal grabber, so could be replaced with 900. (think so)

Issue #28, opened 2016-06-20
Priorities enhancement


penfold42:
Move all default priorities to a single default-priorities.h

Enhance Json priority list with the name of which components have set that priority

Log a warning of priorities collide so it's more obvious why it's flickering as they fight to the death

brindosch:
> Enhance Json priority list with the name of which components have set that priority

of couse, tihs list is currently near to useless. Just if you really know what is going on you understand it.

every device that set a priority should send a identifier? Kodi grabber addon for example "HyperionKodiAddon" or so. Or if a smartphone app is used "HyperionAndroidApp"

penfold42:
ideally, yep

I've uploaded a list of every file under src and libsrc that has the word priority and a 3 digit number in it
[priority-grep.txt](https://github.com/hyperion-project/hyperion.ng/files/323467/priority-grep.txt)

redPanther:
I wish a defaults.h with all defaults not only the priorities.

problem with prios is, there is no one fits all solution. It always depends on what you want to achieve.
Now udpListener(700) is more important as boblight(900) - why?
It is ok when 2 inputs had same prio, as long as they don't send at the same time.
When working with multiple remotes (phone app, web) then a not wisely setted app prio can have funny effects.

brindosch:
usually people would not change these priorities as long they don´t need it. So it is in our hands to set the right values for most common cases.

redPanther:
> Enhance Json priority list with the name of which components have set that priority

then every setcolor/setImage has to be extended. Perhaps with an id that qualifies the sender.
The sender id can be stored in InputInfo struct of prioMuxer. 
before setcolor/setimage can be used, the component has to be register it's id to hyperion class. Then we can store a human readable string, that we can show in json list

redPanther:
> Move all default priorities to a single default-priorities.h

instead of this we can add default values in json schema file.
... this is easier to maintain and get an overview over the values.
Exception for now are the leddevices. There we need some other extension for schema definition and checking. More thoughts on this later.

redPanther:
from the 3 things mentioned in issue, last 2 things already implemented, The first things we do via json schema files. Sadly we can't get the default values in code, but we will get this work to ...

I would say, main objectives of this issues are full filled. I suggest a close.
(we can make a nue issue "set defaults form json schema" ... )

penfold42:
Happy to close.
[...]

Issue #14, opened 2016-06-19
hyperiond and crash status


brindosch:
If we want to move forward to a webinterface and configuration in general we need to ensure that you are able to fix issue with the webinterface as much as possible.

If hyperion leads to a crash the websever is also gone and you just get a blank page (?)
Parts that could lead to this
- a corrupted .json or a not compatible one
- wrong configured led device (output f.e.)
- a enabled component that is not found at the system (v4l2)

we should find ways to prevent here a crash of hyperion that you are able to fix it. (As long that could done) and create better logoutput that describes better what happend now.

redPanther:
when loading json fails, hyperion shouldn't abort, just don't start the whole hyperion stack - only webserver is started. This can be achieved with the new hyperiond structure ...

No component should hardly exit hyperion - just throw an exception that we can catch in main.cpp and react.

brindosch:
Hm, a corrupted .json. It is for sure possible to fall back into a emergency state but this means the json checker needs to fix the confguration itself (which may be hard to code)? We could also add a ".json corrupt - copy and replace with json.default and restart"

redPanther:
when checker fails, do not load given config, use default values instead. (or our default json file).
Then we can modify the config via web ui.

brindosch:
yes, i think restore the .json.default is a easier solution. As we don't need no additional code to maintain or extend, just the checker.
But there is also one point i still not thought about:
What happend if you push later a new release and of course, the old configs don't contain the new array? So than we need to write this array.

redPanther:
I will look which options still have no default and leads to abort. Giving all json  param a default, means we can start without a configfile

brindosch:
//config migrator missing

Issue #6, opened 2016-06-16
new "idle state" with effect or color


brindosch:
https://hyperion-project.org/threads/split-booteffect-and-static-color-add-color-effect-as-new-idle-state.145/

brindosch:
I still had no time to test it. But we need to check all things later anyway.

PR #30, opened 2016-06-20
rework bootsequence

redPanther:
rework bootsequence settings and behaviour.

others:
a small fix in hyperiond: dispmanx grabber was created twice
boblight/udplistener now uses hyperion::getInstance
webserver is alive when daemon crashes (atm not in every situation, because of evil exit() commands in e.g. leddeviceFactory)

yes :-)
bootsequence block is gone, new initialEffect should be inserted:
```
	"initialEffect" : 
	{
		"background-effect" : [136, 97, 50 ],
		"foreground-effect" : "Rainbow swirl fast",
		"foreground-duration_ms" : 3000
	},
```
You can set a color RGB value or a string with effect name. There is no priority settings anymore!!!
Foreground effect/color is set to highest prio (0) as the where in original tvdzwan code. Background effect/color is set to lowest possible prio (32bit systems: 2^32-1 64bit systems: 2^64-1).

Foreground effect is always shown. Background effect/color is only visible if no other input source is sending data. 

It is also possible to modify effect args (this is not a new feature, but name has changed)
```
	"initialEffect" : 
	{
		"background-effect" :  "Rainbow swirl",
		"background-effect-args" : { ... the effect args ... },
		"foreground-effect" : "Rainbow swirl fast",
		"foreground-effect-args" : { ... the effect args ... },
		"foreground-duration_ms" : 3000
	},
```

#28 #14 #6

### DIFF
--- a/include/boblightserver/BoblightServer.h
+++ b/include/boblightserver/BoblightServer.h
@@ -25,7 +25,7 @@ public:
 	/// @param hyperion Hyperion instance
 	/// @param port port number on which to start listening for connections
 	///
-	BoblightServer(Hyperion * hyperion, const int priority, uint16_t port = 19333);
+	BoblightServer(const int priority, uint16_t port = 19333);
 	~BoblightServer();
 
 	///
--- a/include/udplistener/UDPListener.h
+++ b/include/udplistener/UDPListener.h
@@ -25,7 +25,7 @@ public:
 	/// @param hyperion Hyperion instance
 	/// @param port port number on which to start listening for connections
 	///
-	UDPListener(Hyperion * hyperion, const int priority, const int timeout, uint16_t port = 2801);
+	UDPListener(const int priority, const int timeout, uint16_t port = 2801);
 	~UDPListener();
 
 	///
--- a/libsrc/boblightserver/BoblightServer.cpp
+++ b/libsrc/boblightserver/BoblightServer.cpp
@@ -5,9 +5,9 @@
 #include <boblightserver/BoblightServer.h>
 #include "BoblightClientConnection.h"
 
-BoblightServer::BoblightServer(Hyperion *hyperion, const int priority,uint16_t port) :
+BoblightServer::BoblightServer(const int priority,uint16_t port) :
 	QObject(),
-	_hyperion(hyperion),
+	_hyperion(Hyperion::getInstance()),
 	_server(),
 	_openConnections(),
 	_priority(priority)
--- a/libsrc/udplistener/UDPListener.cpp
+++ b/libsrc/udplistener/UDPListener.cpp
@@ -10,15 +10,14 @@
 #include "utils/ColorRgb.h"
 #include "HyperionConfig.h"
 
-UDPListener::UDPListener(Hyperion *hyperion, const int priority, const int timeout, uint16_t port) :
+UDPListener::UDPListener(const int priority, const int timeout, uint16_t port) :
 	QObject(),
-	_hyperion(hyperion),
+	_hyperion(Hyperion::getInstance()),
 	_server(),
 	_openConnections(),
 	_priority(priority),
 	_timeout(timeout),
-        _ledColors(hyperion->getLedCount(), ColorRgb::BLACK)
-
+	_ledColors(Hyperion::getInstance()->getLedCount(), ColorRgb::BLACK)
 {
 	_server = new QUdpSocket(this);
 	if (!_server->bind(QHostAddress::Any, port))
--- a/src/hyperiond/hyperiond.cpp
+++ b/src/hyperiond/hyperiond.cpp
@@ -5,6 +5,8 @@
 #include <QLocale>
 #include <QFile>
 #include <QHostInfo>
+#include <cstdint>
+#include <limits>
 
 #include "HyperionConfig.h"
 
@@ -62,7 +64,7 @@ HyperionDaemon::~HyperionDaemon()
 
 void HyperionDaemon::run()
 {
-	startBootsequence();
+	startInitialEffect();
 	createXBMCVideoChecker();
 
 	// ---- network services -----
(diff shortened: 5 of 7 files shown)


brindosch:
-> https://www.youtube.com/watch?v=bWcASV2sey0
:heartbeat:

Appreciate it! Really! This is great and open so many possibilities!

redPanther:
you made my night ;-)

brindosch:
Now we may have a little problem, how could i disable the single options? leave it empty? This breaks a little with the idea to provide a "enable" option for every option (also if i think about a -always the same- usage from the jsonwriter. You need to think about a workaround. Also to disable/enable the component(s) during runtime may be a problem.

But i trust you :)

redPanther:
I can provide enable if whished for the whole block. Don't no if it make sense.
If you didn't wan't e.g. background-effect you can:
- remove the line
- comment it out ///
- set to empty value ""
- set to black [0,0,0]

choose you way, the result will be the same. Same for foreground, except if you set black, then x ms it is dark on startup.

It is not possible to enable/disable it on runtime, because this is no running service like a grabber or udplistener. But of course this settings will be editable from webconfig - but therefir enable flag isn't needed.

brindosch:
My wish is a clean solution, if you want to disable via config writer a setting: just set the flag to disable and initiate a restart.

I thought this was one of our targets? Usage (writing) for all components the same?
You need a workaround if you need to set empty values.

brindosch:
```
    "foregroundEffect" : 
    {
        "enable" : false,
        "foreground-effect" : "Rainbow swirl fast",
        "foreground-effect-args" : { ... the effect args ... },
        "foreground-duration_ms" : 3000
    },
    "backgroundEffect" : 
    {
        "enable : true,
        "background-effect" :  "Rainbow swirl",
        "background-effect-args" : { ... the effect args ... },
},
```
This would be easy to "call"
foregroundEffect enable true
backgroundEffect enable false
for example

redPanther:
lots of config blocks a related to particular classes in hyperion. When config is not there, then the object from that class is not created. This is a bit weird when other objects want to use that particular object. They have to implement tests and when there is something wrong we get a null pointer exception to runtime.

idea is, all objects are always there, but the can be active and inactive.

Lets take the v4l2 grabber:
the grabber has already start() and stop() methods. With this method we can switch it on and off.

current state: when no config block v4l2-grabber exists, then the v4l2 object is not created and therefore start/stop can't be called during runtime.
If we had a button on remote app/web ui "enable v4l2 input" we had to call v4l2 start/stop - therefore we must have a living v4l2 object!

The "enable" setting enables use to do that. On startup when enable true then create grabber and start it.
If enable false or config not there, v4l2 grabber is created, but it is in a kind of standby state. Grabber is unconfigured and waits that somebody give the v4ls config and call the start methode.
Because v4l2 is not started, it doesn't send anything and don't interfere with other input sources.

Then from somewhere (phone/web app) it gets a start command. Then grabber rereads his config settings and if they ok, v4l2 will switch into active mode and constantly sending frames.

for grabber idle detection:
when grabber is active, get e.g. every 5 seconds a frame. when the frame matches idle condition (e.g. a certain color with a variance and threshold for x seconds) it switches to idle() mode. This means nearly the same as calling the stop() methode, except grabber continues taking constantly screen shots for checking if idle condition is still fulfilled. If not switch back to active mode and send the frames.

The initalEffects are somehow different. They not related to particular objects. This is more a feature for lazy people.
You achieve the same thing when set effect/color to channel 0 and to channel 2^16 with andoid app/hyperion-remote ...

The good question is why we should have an enable option there?
It doesn't harm and is good to see if leds working. When configure the initialEffect effect via webUI there could be a select box with all effects available and "static color". We don't offer a on/off switch, if somebody want's darekness he/she had to set it to black.

redPanther:
BTW I see the enable for boblight/udplistener is not implemented in the right way. This services should have a start/stop method and a constructor that don't start the network service.

brindosch:
> current state: when no config block v4l2-grabber exists, then the v4l2 object is not created and therefore start/stop can't be called during runtime.

Maybe i need to clear this a little bit. You need for v2 a new .json. ALL available options /arrays are available! This is mandatory for startup! So throw away all old .jsons and think from the beginning :)

> for grabber idle detection

Yes, that would be great :)

>  This is more a feature for lazy people.

I work now for months on laziness, but this is what a project makes more and more useable. Find a way to provide features and implement them in a way everybody could use them without to think to hard about a setting. There are always 2 parts: developer use a command line and compile a project, and a user want a easy to use interface and just click the install button. So where should we go in general?

> When configure the initialEffect effect via webUI there could be a select box with all effects available and "static color".

This was my purpose so far

> We don't offer a on/off switch, if somebody want's darekness he/she had to set it to black.
Users view:
--> You explain it and people are lazy and don´t read/understand it and ask you how they could disable it. Cause there is "no option" -->> You show the people something different as you perform at the code side, which needs a workaround for the json writer

Not everything makes sense from the developer view. This may be more often, but from a user view you expect "one configuration flow that behaves the same - feels consistent"

Regarding the start stop method, if i have a working example i could try to c/p it to other components to help as much as possible :)
The same for the logging rework i could do it, but need a short example how i implement it the right way. We should write it down for documentation anyway.

brindosch:
@redPanther 
I tested it and it´s so great!
:heart:

brindosch:
I found a issue. If i place "Warm mood blobs" it starts the "cold" blobs. This affects background and forgeground, could you confirm this?

redPanther:
the default config has empty "...-effect-args" this causes the effect to use default values. The args param should only be used when you want to change param of an effect. The params are different per effect, you have to look into effect python file to see what params possible. This is, if you want to override params in initial effect. (e.g. fade from black to startup color.)

I suggest to comment out params in config. BTW this behaviour was there before I made any changes in this code, but there was never an empty args param in config file

brindosch:
Ah, yes you are right, we should remove it entirely from the default file.
I am sorry

PR #34, opened 2016-06-21
logger: set global log level

redPanther:
Add ability to set loglevel for all loggers and for specific loggers.

Example:
```
// will set min log level for all loggers to warning
Logger::setLogLevel(Logger::WARNING);

// will set loglevel of specific logger to warning
// if global log level is set (above) this is useless
Logger::setLogLevel(Logger::WARNING, ”EFFECT”);
```

Next step is to connect this with config ....
No

### DIFF
--- a/include/utils/Logger.h
+++ b/include/utils/Logger.h
@@ -25,9 +25,11 @@
 class Logger
 {
 public:
-	enum LogLevel { DEBUG=0, INFO=1,WARNING=2,ERROR=3 };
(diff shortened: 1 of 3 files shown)


brindosch:
Every day one step :)

penfold42:
Can we over ride the debug level from the command line ?

redPanther:
Not yet, but we will. Atm you have to set the level in msin.cpp
I like the level WARNING than all ongo is gone.

PR #44, opened 2016-06-24
logging migration for ws281x and LedUdpDeviceLeddevice logger

penfold42:
logging migration for ws281x and LedUdpDevice


brindosch:
one small thing: as you see, your branch leddevice_logger is based on master which you pushed already your other PR. So both PRs contain now the the cmake compile warning fix. You should consider to push changes to your master just if you don´t plan to open more than one PR at the same time :)

PR #51, opened 2016-06-25
add .travis.yml

brindosch:


brindosch:
note: I will add deployment soon

redPanther:
Can we use a script instead of chained commands?

insteda of:
```
script: mkdir build && cd build && cmake -DPLATFORM=x86 -DCMAKE_BUILD_TYPE=Release .. && 
make -j 2
```
e.g. this:
```
script: bin/travis_ci.sh
```

BTW it is good to use `make -j $(nproc)` instead of `make -j2`. Then we use always maximum available cpu cores.

can you also add a
`make -j$(nproc) package`
then the testbuild also tests "make install" and package generation - this will show if something wrong in cmake. For Testbuild I suggest following cmake args:

`cmake -DPLATFORM=x86 -DCMAKE_BUILD_TYPE=Release -DENABLE_AMLOGIC=ON -DENABLE_TESTS=ON -DENABLE_SPIDEV=ON ..`

to get maximum possible code coverage

penfold42:
Can we also do arm tests ?
https://www.tomaz.me/2013/12/02/running-travis-ci-tests-on-arm.html

redPanther:
I would also vote for that :-)

brindosch:
Feel free to create a PR :)
You may faster  in setting this up :)

redPanther:
I'm not realy satisfied with the posibilities mention in the link to compile for arm. I think our toolchain is better. We have to setup a good crosscompile image ....

brindosch:
On the other hand i hate crosscompile. It throws sometimes false positive (errors) and i was not able to compile with zeroconf enabled (all libs available!).
Maybe i have done something the wrong way, no idea. But it really bothers me often.

penfold42:
I was looking at OSMC sources yesterday.
They do a whole qenu chroot build too - it took AGES !

Sam's view seems that the performance hit was worth it to avoid the cross compile hell

redPanther:
with zeroconf there is a problem atm. The library names are hardcode atm. We need a cmake "find_package" instead. E.g. current hyperion won't compile on distro's with no static avahi libs. (e.g. archlinux). Therefore I added a flag to use dynamic avahi libs ... 
I need to get the time to install raspbian (normaly I use arch on my pi's)

penfold42:
Ditto centos x86

Issue #25, opened 2016-06-20
migrate logging to new framework


penfold42:
Lots to be done here - if someone new to hyperion wants to contribute, this is a tedious but worthy task!

There are 178 std::cout calls in the code as show here:
/hyperion.ng# find src libsrc -type f -exec grep -n std::cout /dev/null {} \; | wc -l
178

brindosch:
no problem, i just need a code pattern /howto that could be used from me and for docu that devs could understand it.
regarding include, loglevel variants and how i call them

redPanther:
please also take printf, std::cerr and qDebug() in account ;-)

redPanther:
info for current head of master regarding new logger:
you can add commandline options for loglevel:
view debug messages:
```
hyperiond --debug
```
(more see help)

redPanther:
Call for help! I wrote some doc about logging in the wiki:
https://hyperion-project.org/wiki/Logging

Would be great if somebody can migrate logging. We don't need migrating logging at once - component by component is absolutly ok. Ifquestions ... I will assist

brindosch:
Good idea, i will have a look. For sure i could do that boring part :)

redPanther:
Starting with the blackborder files are nice to practice

penfold42:
What about banning commits of files that don't use the new logging framework ?

brindosch:
Not sure what you mean in specific. Remove all code which has no logger implemented? This break Hyperion?

redPanther:
Dont accept pr's when new logger oisnt used

brindosch:
Of course.

brindosch:
I have a question regarding the GetInstanceOf. This result in? another logoutput? a phrase like "EffectEngine" before the message itself?

penfold42:
a new logging object if it doesnt exist.

If it does, you get a handle to the existing logging object

brindosch:
This results in: you no longer see which part of Hyperion has thrown this error? If the message itself contain no "identifier"

penfold42:
Sorry, I missed part 2 of your question.

The name of the logging object IS printed

redPanther:
try to explain:

first get a logger:
`log = logger::getinstance("effects")`

this will create a logging object with logging scope "effects", if not exists, otherwise it returns existing object.

on log messages:
Info(log, "my msg")

[...]

PR #52, opened 2016-06-25
Logging leds

redPanther:
migrate leddevices to new logger.

still todo:
LedDeviceWS2812b
LedDeviceWs2812SPI

#25

### DIFF
--- a/include/leddevice/LedDevice.h
+++ b/include/leddevice/LedDevice.h
@@ -7,6 +7,7 @@
 #include <utils/ColorRgb.h>
 #include <utils/ColorRgbw.h>
 #include <utils/RgbToRgbw.h>
+#include <utils/Logger.h>
 
 ///
 /// Interface (pure virtual base class) for LedDevices.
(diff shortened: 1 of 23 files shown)


brindosch:
Nice!
There is something strange, this PR has partly the same code change like the previous. How could this be?

penfold42:
Because he has done it more completely for more led device drivers.
He also instantiated the logger at the parent class which is much cleaner!

redPanther:
I made the pr from the same branch as Iusef for my prior pr. This pr is just in addition.

After you merged. I forget to git fetch and merge to sync our repos. Thatswhy github shows my prior and new commits .instead of new one only.
Thats not harmful, because git detect that during merge and do the right thing.

BTW thx for settiing up CI. An additional osx build would be cool

redPanther:
... forgott to say, with this comit we can start without configured led device. If no led device configured, then dummy device is selected (dummy means - file device with output /dev/null )

brindosch:
nice on, currently i added the file device with output /dev/null as default. I will do soon a PR to update our current .json and add a new one without comments which should be used if you perform a "reset this configuration to default" at the webinterface :)

PR #53, opened 2016-06-26
UDPListener improvements:

penfold42:
- if you listen to a multicast address, it now also listens to all ipv4 addresses

- shared udp sockets - multiple hyperiond instance can optionally share the same udp packets
False is the defaul

"shared" : true

### DIFF
--- a/include/udplistener/UDPListener.h
+++ b/include/udplistener/UDPListener.h
@@ -26,7 +26,7 @@ public:
 	/// @param hyperion Hyperion instance
 	/// @param port port number on which to start listening for connections
 	///
-	UDPListener(const int priority, const int timeout, const std::string& address, quint16 listenPort);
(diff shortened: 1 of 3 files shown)


brindosch:
Nice one :)
I need to keep track of this for documentation (later)

brindosch:
@penfold42 
a little bit OT:
Regarding the zeroconf/avahi in general.
Lets assume i use the same name on every config (device -> name : MyHyperionConfig). It is not posssible for an app to sort which JSON and PROTO port is now which instance?

We might need a clear Hyperioninstance@Hostname that it never be a user fault and at least the app could make a difference?

redPanther:
... this is already done. hyperiond/main.cpp line 265

penfold42:
Yup - unless the user has 2 instances on the one host with the same device name were good

PR #61, opened 2016-06-27
Added support for sk6812RGBW leds (ab)using the SPI interface.

penfold42:
Added support for sk6812RGBW leds (ab)using the SPI interface. right now it only uses RGB - white is set to zero

 "device" :
        {
                "name"       : "MyPi",
                "type"       : "sk6812spi",
                "colorOrder" : "grb",
                "output"     : "/dev/spidev0.0",
                "rate"       : 3800000,
        }


brindosch:
nice one. Hopefully you find a way to use the CW/WW led.
i thought about something like:

If the values for R/G/B are near the same (which is by defintion white), white could be enabled. Now depending on average val (0-255?) the brightness of CW/WW could be set.

OT1: If i imagine right the SPI interface needs a (real)root class to remove code duplication (like RS232). @redPanther : yes? maybe i missed this.

OT2: I am highly interested at your clock effect, may we see a PR soon? :)

redPanther:
> OT2: I am highly interested at your clock effect, may we see a PR soon? :)
its optimized for my special matrix lamp. Have to make it general before releasing

penfold42:
That white algorithm is already implemented in the 281x based device and I need to port that (and other alohirthms) to this device.

How would a new root class look and feel compared with the existing LedSPIDebice class that this uses ?

PR #64, opened 2016-06-28
fix config install in deb package

redPanther:
the default config was renamed - this produced an error in debian postinstall. @brindosch please change config name in cmake/debian/postinst if you change the config name again.


brindosch:
Oh, i´m sorry!

PR #70, opened 2016-06-28
travis.ci: improve build options, add osx builds

brindosch:
OSX DISABLED
OSX won´t run currently. I need help to get a actual list of dep names.
the build still fails with qt5 not found. looks we need to tweak more, cause qt5 is installed. maybe not linked.


brindosch:
There is still a error not sure of this is fixed with the pr of @penfold42 
We walking about arguments, can´t be that hard :)

redPanther:
of course penfolds pr is needed for that ...

PR #72, opened 2016-06-29
Update travis_install.sh

redPanther:
osx dep update

add osx cmake dependency


brindosch:
the build machine said that cmake is already installed, but it should not harm.

We need the cmake path for qt5, i set it now manual for osx, but this will break soon cause its more a hack but its working as a first test. For sure these is a cleaner solution available.

There is something which i did wrong, is the make command wrong? why it starts no compile after configure?
https://travis-ci.org/brindosch/hyperion.ngBeta/jobs/141262602
you could see the cmake path i added if you click on the commit diff at the top.

Issue #68, opened 2016-06-28
Device { leds } parameter should work with any device type


tociek:
Related to #626.

As requested, opening an issue here... 

"leds" parameter in device section works only for apa102 leds. Ideally, it should work for any device type

redPanther:
why does it make sense? the count of leds are the count of configured leds in the leds block. I think we should remove this parameter from apa102 device. can somebody point it out why apa102 needs thos special treatment?

tociek:
Normally you can achieve this effect by defining these extra leds that you have, but you want them to be off - and assign special transformation to these leds with brightness=0. So currently it is possible but  a bit cumbersome to do. "leds" parameter sounds like something convenient to achieve this effect.

Problem with "extra" leds being on is not only with apa102, but with ws2801 as well. I don't know what happens with ws2801 but it has been a problem with hyperion since I remember (at least for 3 years)

penfold42:
The effects engine needs to know how many LEDs it should be using - this is why we need the concept of maximum physical LEDs bs maximum configured/usable LEDs.

I agree that doing it the led devices is the wrong place.

Maybe the transform or even just before the call to updateleds is a much better place

tociek:
While I agree that doing this in each led device code would be "suboptimal", however from config stand point "device" section does make sense as "leds" describes physical device (and not transformation)

penfold42:
Agreed - where it is configured doesn't have to align with where we implement it.

redPanther:
this should be done in hyperion class. today my day is a bit tight. tomorow evening I have time. I have a nice idea to solve this and other related issues. more later ....

penfold42:
I just remembered that the rpi_281x driver needs to know how many physical LEDs there are at startup time.

redPanther:
this issue is solved. see pr #77 for details

PR #77, opened 2016-06-30
common ledbuffer for color transform

redPanther:
hyperion class uses a common buffer for all operations on ledColors got from muxer
all color transforms uses new ledBuffer instead of making copies of ledbuffer

remove colorCorrection, because it does 100% of colorTemperature
(side note: color transformations need definitly a rework! )

#68 add ability to define amount of real hardware leds (can be more then configured leds)
Those leds are set to black/off

new optional config option (if not needed, don't add it to config)
```
"device" :
{
  "ledCount" : 120
}
```

other fixes:
fix compile bug in profiler
update doxygen config
migrate logging for hyperion and color transform classes (#25)

### DIFF
--- a/.travis/travis_build.sh
+++ b/.travis/travis_build.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 # compile hyperion on osx
 if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
-	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON .. && make -j$(nproc) package
+	cmake . -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt5/5.6.1-1
+	mkdir build && cd build && cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON -Wno-dev .. && make -j$(nproc) package
 fi
 
 # compile hyperion on linux
 if [[ $TRAVIS_OS_NAME == 'linux' ]]; then
-	mkdir build && cd build && cmake -DPLATFORM=x86 -DCMAKE_BUILD_TYPE=Release -DENABLE_AMLOGIC=ON -DENABLE_TESTS=ON -DENABLE_SPIDEV=ON .. && make -j$(nproc) package
+	mkdir build && cd build && cmake -DPLATFORM=x86 -DCMAKE_BUILD_TYPE=Release -DENABLE_AMLOGIC=ON -DENABLE_TESTS=ON -DENABLE_SPIDEV=ON -Wno-dev .. && make -j$(nproc) package
 fi
--- a/bin/service/hyperion.systemd.OE.LE.sh
+++ b/bin/service/hyperion.systemd.OE.LE.sh
@@ -1,0 +1,12 @@
+[Unit]
+Description=Hyperion ambient light systemd service for OpenELEC/LibreELEC/RasPlex
+After=graphical.target
+
+[Service]
+ExecStart=/bin/sh -c "exec sh /storage/hyperion/bin/hyperiond.sh /storage/.config/hyperion/hyperion.config.json"
+TimeoutStopSec=2
+Restart=always
+RestartSec=2
+
+[Install]
+WantedBy=default.target
--- a/bin/service/hyperion.systemd.sh
+++ b/bin/service/hyperion.systemd.sh
@@ -1,15 +1,15 @@
 [Unit]
-Description=Hyperion Systemd service
+Description=Hyperion ambient light systemd service
 
 [Service]
 Type=simple
-User=root
-Group=root
-UMask=007
+User=hyperion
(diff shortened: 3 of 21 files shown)


penfold42:
looks good - I was just staring at that very section of code earlier.

I was looking at the color order and how we might use the distinction between physical and defined leds.

I're really like to move the color order code out of here - it does my head in working on the led device drivers getting confused by red possibly not being red at the driver level!

redPanther:
my plan is (in short) that _ledBuffer has size of hwleds. all color filters operate on size of configured leds. ... this is basicly how your stuff worked, but on a more general level. 
Additionally (not yet but soon) I want to move filters to separate a filter engine

penfold42:
Perfect

What are your thoughts on the colour order ?
In my mind it's an led device thing so could be implemented there where required

redPanther:
color order: make it as a function and then put it into device ... every device has colorrgb as input.
It's a bit like the grabbers. not all grabbers are rgb. the transform it in there space. If we made a nice func for that also grabbers can use this func

penfold42:
This will help with rgbw too.
I already wrote a crappy RGB to rgbw transform - we could
Have a remap colours function that does the color swap as well as the rgbw conversion

redPanther:
implement changes for #68 description comes later ...

redPanther:
Please wait with merge, have to do some cleanup.

redPanther:
While migrate to new logger, I found some strange thing about color/temperature  correction.
https://github.com/hyperion-project/hyperion.ng/commit/257532df74d0842af8fca3cb2122869e8b7eeb71

This issue have to resolved first, before I can go on ....

brindosch:
the whole color correction needs a cleanup too.
drop old values and just use the new ones.

But there is still something "strange" with HSL luminanceGain VS HSV gain.
If i lower my HSV gain from 1.0 to 0.5 it is much better than HSL luminanceGain from 1.0 to 0.5
The colors looks very over saturated with HSL, am i the only one who still use the deprecated HSV gain here?
@AEtHeLsYn

penfold42:
I still do !

I have a led strip for mood lighting and my current script used V.

It wouldn't be a tragedy If we lost it - Its probably better I change the RGB value when I set the colour rather than fiddle a transform value

redPanther:
After some investigation I found out that correction and temperature are the same. I will delete correction, because it is obsolete. I also looked through the other 'correction' code - This code needs definitely a rework. I will do this soon.

redPanther:
Did anybody found out what the intention of rgbchannelcorrection.cpp is? Reading the code seriously I got 2 possible meanings. 

1. Since correction values can from -255 to 255 I assume the correction values are for increase/decrease current color. Sadly the functions for getting the correction value are uint8 and will never return negative values. This is not good for decrease current color

2. In applyCorrection you see the correction is nothing else as setting an upper limit for color value and make a linear mapping between input color 0-255 to output 0-limit. (Nothing to do with temperature of color).
The usage of this class is exactly that (multicolorcorrection.cpp), but this doesn't make sense with 1.

There is a problem, this is obvious. To fix it I have to know what the code should do. Both meanings are usable.

redPanther:
ready to merge, I updated my text above ..

brindosch:
affects hyperion-remote.cpp (color correction)?

redPanther:
yes, forgott to remove it. it doeas absolutly the same as temperature

brindosch:
is this not also part of it?
https://github.com/redPanther/hyperion.ng/blob/c38c06d8274fb9be7460542632d61acecf646715/src/hyperion-remote/hyperion-remote.cpp#L170-L181
All the correction args up to the info description

redPanther:
correction still exist in remote, but sends temperature command. This should be ok for now.
I will rework that next week. I have to get into adjustment code before. It looks like there are more duplications ... (and hopefully not more strange things going on)

brindosch:
Hope so!
It is not funny to clean this up.

redPanther:
Had a quick look. It looks like temperature aka correction is already part of adjustment. But as I said I had to read a bit code first.
gn8

penfold42:
Is there some existing colour adjustment code we can borrow ?

Maybe from a media player like vlc or even from an open source tv !?

redPanther:
Perhaps, but integrating foreign code is often more work than implementing it. We have the structure to do what we want, but 
We should define what we want to achieve. 

Our code is ok, needs cleanup and so on ...

I think biggest problem is the user. Having lots of color correction possibilities makes not easier for the user, albeit I realy like to have endless params for everything.

As a quick idea, we need a phone app that takes video frames and set calibration pictures and try settings until best match is found. Or a hyperion app take screenshots via v4l2 webcam and calibrate

penfold42:
That would be cool - and as long as the image is of reasonable
Quality who cares how imbalanced it is - if you tell the code the are with the tv test pattern and the led wall illumination it could probably work out the rest from there to get them to match

PR #94, opened 2016-07-10
Update README.md

NicoHood:
Not sure if we should move the compile instructions completely to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,30 @@ HYPERION
 This is a (alpha) development branch for the next major version of hyperion
 
 
+
+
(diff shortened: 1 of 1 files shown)


brindosch:
I suggest to update the readme at non ng repo. This repo should not be used for daily usage. I don't want that people open more and more issues here that is a alpha problem.

Paulchen-Panther:
Was ist eigentlich mit diesem PR? Kann das weg? Oder?

redPanther:
changes made in #306

PR #98, opened 2016-07-11
implement auto detection of used platform. 

redPanther:
to minimize errors while compiling, now the PLATFORM can be autodetected.
If you don't add -DPLATFORM to cmake command, then it tries to guess the platform.
It works very well for x86 and apple. The RPI stuff is not tested atm. I will test it later, but if somebody can test it in the mean time, this would be great.
Don't forget to wipe your build directory to avoid conflicts with prior builds (cd build && rm -rf *)

if you do not want auto detect, then add -DPLATFORM as usual. (and again, when playing/testing this feature, wipe your build dir before)

no


brindosch:
Nice :)
checked
it uses rpi-pwm as it should

redPanther:
I will check it on my pi1, there "rpi" should be used (as I understand the pwm stuff right?)
wetek/amlogic is straight forward and should be working as expected

brindosch:
you can't differ between pwm and non pwm. At least i don't know how the system should tell you that. This is the reason why i use always rpi-pwm for release. There is no non pwm version available, cause this needs user interaction i want to avoid. No question, no error :)

brindosch:
pwm is available for all Pis? So it works as it should. You could disable the ws281x for pi1. But that's it.

For a release version we keep to simplify the process everything on.

redPanther:
good to know, I will tune cmake to minimize the options

brindosch:
Hm, no you can't disable ws281x on Pi1, if @penfold42 adds the support for Pi zero revision 3 (which is a Pi 1 if you search for the BCM version (see install_hyperion.sh)) we just run in a complicated documentation.

brindosch:
I still vote for a removal of the pwm options and merge them to one pwm device. Easy to understand and working for everyone - "hardware" handling should be done from hyperion and not from the user. @penfold42

penfold42:
@hyperion-project should fork https://github.com/jgarff/rpi_ws281x/tree/sk6812?files=1

Apart from needing HDMI plugged in it seems to work for all models of pi

brindosch:
your wish is my...
https://github.com/hyperion-project/rpi_ws281x

If @redPanther ou you could finally test this (ws281x device) with ws2812b on a Pi1 would be great. So we could remove the ws2812b device and use this is general solution.

redPanther:
Currently I'm collecting parts for better hyperion testing. I've got an 'old' pi2 from somebody for my packaging and general rpi test. Just in that moment I got some lpd leds from a technician at work. Now I try to get a few 2801 leds and I have some meters spare ws2812 + several arduinos. My goal is to get several hardware setups ready to use without tinkering around.

I can test ws2812, but I have to make the right cables - but for my goal I have to do this anyway

brindosch:
I also thought about a test station :)
Sadly i have not much space to setting up something like this :/

penfold42:
I am tempted to make a multiplexor to rake PWM, SPI and USB dongle outputs from a couple of Pi's to automate the hardware testing

redPanther:
I take only a few leds (8leds or so) and put them onto the case of my pi. My plan is to put one spi and one pwm on the same pi. Then another mini box with an arduino and also some leds onto the case for adaligt/serial port tests. My goal is not a full tv setup or so, just a small box to test function 

@penfold42 cool :+1:

penfold42:
On a related note, do you know how to build for pi1 using a pi2 ?
I've forgotten how slow the pi1 is !

redPanther:
the bins for pi2 should work on pi1 to, as long as you use same distro version (dependencies ... ). If both use e.g. jessie ... no prob.

I compile always on my desktop and when all things working I start compiling on pi - without any hurry  - and doing other usefull stuff in the mean time ... e.g. household ;-)

brindosch:
could you explain me why you are forced to build on a Pi1?
Hyperion runs on all Pis, or just for testing?

penfold42:
not true.... instruction set is slightly different between them.
Pi1 exe will work on a Pi2/3, but the reverse is not guaranteed

brindosch:
To cool this down, Hyperion have been downloaded 180 times a week for Pi1, see my statistic workaround to get this numbers :) They are all the same. (An Pi3 still counts as Pi2)
I never read about such a problem.

https://sourceforge.net/projects/hyperion-project/files/release/

penfold42:
Here's what I get building hyperiond on pi3 osmc and then try to run it on a pi1 osmc:

osmc@osmc:~$ ./hyperiond
Illegal instruction

brindosch:
Regarding the crosscompile and if we could trust the firmware path we use bcm2708 (pi1). I don't know if this is right. All i could say that no one ever complained about this.

penfold42:
The binary packages that are on sorceforge. how are they build and then packaged?

brindosch:
crosscompile on a ubuntu 14.04 vm with raspbian wheezy sources. The (pi).tar.gzs are all the same, i just named them differently. Or what you mean with packaged?

penfold42:
thanks - that'll be why... I bet the toolchain on wheezy builds armv6 binaries

PR #103, opened 2016-07-11
Logging for dispmanx, framebuffer, x11, v4l2, osx

brindosch:
Logging for dispmanx, framebuffer, hyperion-x11, v4l2, osx
no
https://github.com/hyperion-project/hyperion.ng/issues/25

### DIFF
--- a/include/grabber/DispmanxFrameGrabber.h
+++ b/include/grabber/DispmanxFrameGrabber.h
@@ -11,6 +11,7 @@
 #include <utils/Image.h>
 #include <utils/ColorRgba.h>
 #include <utils/VideoMode.h>
+#include <utils/Logger.h>
 
 ///
 /// The DispmanxFrameGrabber is used for creating snapshots of the display (screenshots) with a
@@ -84,4 +85,6 @@ private:
 
 	// size of the capture buffer in Pixels
 	unsigned _captureBufferSize;
+	
+	Logger * _log;
 };
--- a/include/grabber/FramebufferFrameGrabber.h
+++ b/include/grabber/FramebufferFrameGrabber.h
@@ -5,6 +5,7 @@
 #include <utils/ColorRgb.h>
 #include <utils/VideoMode.h>
 #include <utils/ImageResampler.h>
+#include <utils/Logger.h>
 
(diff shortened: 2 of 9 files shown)


redPanther:
this is not optimal:

```
Info(_log, "Cropping from image (width x height) %dx%d", _width, _height);
Info(_log, " left: %d", cropLeft);
Info(_log, " right: %d", cropRight);
Info(_log, " top: %d", cropTop);
Info(_log, " bottom: %d", cropBottom);
```
... because it splits coherent inf accross logger statements. 
this contains same info and is in one statement:
```
Info(_log, "Cropping image: width=%d height=%d; crop: left=%d right=%d top=%d bottom=%d ", _width, _height, cropLeft, cropRight, cropTop, cropBottom);
```

brindosch:
Ah, yes i know. i tried it first with newline, but this looks ugly. Is there no better solution available?

redPanther:
logging is no ascii art. If you want to put a message it has to fit in a line.

This is the difference to freely put something on terminal. If we want something nioce for the user -> this is no logging -> do not use logger.
But hyperion is a daemon and it has no ui for the user, so there should be no reason for nice painted terminal art.

brindosch:
:P

redPanther:
... I had countless discussions over that @ work ... drawing big tables in logfiles and then complaining if another thread or the log rotation scrambles the ascii art ... *grrr* and poor people that should made automated processes for log evaluation.

brindosch:
I made a mistake, v4l2 should output a float value. instead i get this with %f
```
[HYPERIOND V4L2GRABBER] <INFO> Signal threshold set to: 1717788323676192129486435861422674573983...
```
Duo the point it uses in master %d which can't or should not work, i missed a convert for this value, could someone point me to the right one?
https://github.com/hyperion-project/hyperion.ng/pull/103/files#diff-aec2cfe446c748ff341302e49e490d6bR80

redPanther:
use `%4.2f` you can play with the numbers to get a result that satisfy you
http://www.cplusplus.com/reference/cstdio/printf/
(precision)

Issue #111, opened 2016-07-12
Smoothing improvement (small)


brindosch:
@redPanther 
If you may have some time later, there is a small issue now. If you use the backlight option it is much more noticebale.
The smoothing should stop the output not immediately after "no change" -> stop output trigger.
It would be better to keep the output flow -until- the configured smoothing time is reached.

If you have a fast color change: the leds "hang" in a smoothed state between for example blue and black. (If you go from a blue to a black picture)

Thank you

redPanther:
The output is not stopped imidiatly, but perhaps we have to double the time ....

I'm thinking about a qtimer trigger for about 10 seconds + 2* smooth time. 

I'm not satisfied with the modification I've made in the past. For my taste it is better always send in a defined frequency, regardless if there is new data or not. If the leds receives noise from somewhere, the leds would be refreshed and set to its last state. The current implemention wouldn't do that.
Perhaps an option to enable continous led refresh would be the best. I'll have a look ....

redPanther:
whats your smoothing settings? problems occured when updateDelay>0. Hopefully this is resolved now

brindosch:
update delay is 0.
That you delay the output explains why it is more a very light blue, instead something in between (a brighter blue - hald black half blue)

redPanther:
when delay=0 then my latest fix won't work. I disable writing to leds, when smoothing gets no new input.

problem for delay>0 was that the buffer wasn't flushed. e.g. if delay=10, then output stopped 10 frames before input stops. This should work now.

brindosch:
just a small report, i could easy recreate the issue when using kodicheck. If the state switches from background-effect to "grabbing" but the player is in pause mode. You see very good the effect mood blobs. We should enable continous as default.

brindosch:
This is done

PR #105, opened 2016-07-12
Migrate logging and add "enable" to smoothing

penfold42:
**1.** Tell us something about your changes.
Migrate logging and add "enable" to smoothing

**2.** If this changes affect the .conf file. Please provide the changed section
                "smoothing" :
                {
                        "enable"          : false,
                        "type"            : "none",
                        "time_ms"         : 50,
                        "updateFrequency" : 25.0000
                }

**3.** Reference a issue (optional)
#3 #25

Note: For further discussions use our forum: forum.hyperion-project.org

Issue #87, opened 2016-07-10
Segmentation fault


NicoHood:
I compiled hyperion for my raspberry 3 in arch and **sometimes** get a segmentation fault.

```
[alarm@alarmpi3 ~]$ /usr/bin/hyperiond /etc/hyperion/hyperion.config.json
INFO: Creating linear smoothing
HYPERION (CS) INFO: Created linear-smoothing(interval_ms=50;settlingTime_ms=200;updateDelay=0
BLACKBORDER INFO: threshold set to 0 (0)
BLACKBORDER INFO: mode:default
BLACKBORDER INFO: threshold set to 0 (0)
BLACKBORDER INFO: mode:default
PROTOCONNECTION INFO: Connecting to Hyperion: 127.0.0.1:19447
[HYPERIOND MAIN] <ERROR> Hyperion Daemon aborted
QtHttpServer started on port 8099 "Qt Static HTTP File Server"
KODICHECK ERROR: Kodi Connection error (0)
Segmentation fault (core dumped)
```

I only changed the number of leds to 25 and the adalight device:
```json
	"device" :
	{
		"name"       : "MyHyperionConfig",
		"type"       : "adalight",
		"output"     : "/dev/ttyACM0",
		"VID"        : "0x2341",
		"PID"        : "0x8036",
		"rate"       : 115200,
		"colorOrder" : "rgb"
	},
```

SPI is not enabled, nor used. It is not even compiled in this binary I use.

I used this PKGBUILD (which is incomplete, but just for documentation):
```PKGBUILD
# Maintainer: TODO

pkgname=hyperion
pkgver=r1160.0feb1d7
pkgrel=1
pkgdesc="An opensource 'AmbiLight' implementation"
arch=('armv6h' 'armv7h')
url="https://github.com/hyperion-project/hyperion.ng"
license=('MIT')

depends=('libusb' 'python' 'icu')
optdepends=('xorg-server: X11 grabbing')
makedepends=('git' 'cmake' 'qt5-base' 'qt5-serialport' 'avahi')
provides=('hyperion')
conflicts=('hyperion')
backup=('etc/hyperion.config.json')
source=('hyperion::git+https://github.com/hyperion-project/hyperion.ng#branch=master')
sha256sums=('SKIP')

pkgver() {
        cd "$srcdir/${pkgname%-git}"
        printf "r%s.%s" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
}

build() {
[...]

PR #112, opened 2016-07-13
multi fix commit

redPanther:
- refactoring leddevicefactory
- adalight: default for "delayAfterConnect" is 1s now - needed for most arduino's because of there special behaviour on open
- fadecandy: new option for disabling configuration send - if you want to keep your fadecandy defaults
- Hyperion.cpp: simplify createSmoothing discussed in #105
- smoothing:
 - add option for continuous output
 - when updatedelay>0 and continousOutput is disabled, buffer is flushed correctly after no input is detected
- ad doxygen to travis

yes, smoothing gets a new param `"continuousOutput" : false`

#111 #87


brindosch:
nice one
Thank you, i will let you know :)

Issue #107, opened 2016-07-12
Move blackborder config to hyperiond.cpp 


brindosch:
I searched the reason why the enable option for v4l2 won't work, and now i found this
https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/blackborder/BlackBorderProcessor.cpp

should be moved to

https://github.com/hyperion-project/hyperion.ng/blob/master/src/hyperiond/hyperiond.cpp

redPanther:
For my taste the bb'way to hande it is the best. I'm currently updating my fadecandy driver with some things and I do the same.
Another rule for clean code is to avoid very long parameter lists ...

The advantage is: if developer of e.g. bb wants a new config param he_she had to modify only his - bb - files and not the hyperiond.cpp. And he_she had to add less code. Only the class member in the .h file and the line in initializer list.
the "enable" should be implemented inside bb. As addition bb should have a setconfig(const Json::Value &blackborderConfig) function to update class mebers and reinitialize the bbdetector. Then we can create a new config json object in json server and set the new config during runtime.

brindosch:
So kodi should be also removed from hyperion.cpp? The problem is we should consider to implement all (additional?) options the same way. It really doesn't matter how, i am not a devleoper so this is your decision about pro/cons and future functions

redPanther:
let use exercise with leddevices first. look at my fadecandy pr - the setconfig thing and changes in factory.

redPanther:
... should be solved now. enable should work as expected

PR #119, opened 2016-07-15
make "enable" work for blackborder detector

redPanther:
... make "enable" work for blackborder detector
If you look with --debug, then you will see following message in state enabled=false too:
`//[HYPERIOND BLACKBORDER] <DEBUG> <BlackBorderDetector.cpp:26:calculateThreshold()> threshold set to 0.000000 (0)`

bb is deactivated, but the bb detector object is created - but do nothing, just prints that line. If you enable bb, then you will see more messages ....

#3 #107 

### DIFF
--- a/include/blackborder/BlackBorderProcessor.h
+++ b/include/blackborder/BlackBorderProcessor.h
@@ -35,6 +35,18 @@ namespace hyperion
 		BlackBorder getCurrentBorder() const;
 
 		///
+		/// Return activation state of black border detector
+		/// @return The current border
+		///
+		bool enabled() const;
+
+		///
+		/// Set activation state of black border detector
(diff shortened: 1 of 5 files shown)


brindosch:
bb is deactivated, but the bb detector object is created - but do nothing

current state? I know that my BB works with false :)

redPanther:
the trick with "enable" is, that we always create objects, but they don't do any actions - just waiting if something enables this component.

if enable=false bb is definitely off. I moved the config query for "enable" state into blackborderprocessor.cpp, where the other config stuff is. The message comes from the instantiated bbDetecor class in initializer of bbProcessor. To silence the message, I have to pass the enable state to the detector to - but for what, only silence a debug message ...

PR #135, opened 2016-07-21
make avahi optional

stefansaraev:

as title says, avahi is now optional. defaults to enabled


penfold42:
What problems are you having that this will solve ?

In the future this is the expected way for remote control apps and forwarders to discover hyperiond instances (and this will break them)

I know some platforms don't have dynamic library dev packages so we added the static link option

redPanther:
We removed possibility to disable avahi, so I wont reintroduce it. 
If you have probs with static vs shared libs, there is a cmake option to switch between.

What we need is a find_package cmake module for avahi that find correct lib  names, include folder and prefers static libs.

stefansaraev:
> What problems are you having that this will solve ?

none. some of us prefer having as less dependencies as possible.

> We removed possibility to disable avahi, so I wont reintroduce it. 

I'll keep the patch private then, and if needed, may re-add avahi to my embedded distro later :)

penfold42:
Does your "embedded distro" have the OS provided avahi?

stefansaraev:
it is an stripped down OpeELEC fork. no problem adding avahi. I just dont like it ;)

redPanther:
ok, I think we delay this discussion.

brindosch:
There are always to types of peoples. People that want to customize everything or drop functions cause they don't want/need it and people that click on install and are happy.

So the discussion would be endless as there is no real solution. The idea behind is the next step to less configuration (ip searching) and more fun. This is also great for our forwarder as you could auto detect this instances too without any knowledge. Of course there is enought to do. Actually is it useless as no one use it.

Of course we could talk about an option in config.json to disable zeroconf like in kodi for people that don't want or need it.

stefansaraev:
agree with that ^^. 

while I am one of those who want less dependencies, I understand "click on install" is better for most users. you are the ones who write and maintain the code (and take care of bugs) so it is your decision. I wont beg for making avahi optional. it's all good :)

if there is a demand in future (I dont think there will be, maybe 2 or 3 people complain after hyperion.ng goes stable), this can be discussed later

PR #138, opened 2016-07-24
KODIVideoChecker Fix

Paulchen-Panther:
Es sollten jetzt alle zustände (Video, Audio, Picture, Pause, Menu, Screensaver) von Kodi erfasst werden. Auch wenn sich Kodi im Anmeldemodus befindet (Passwort). Der Pause zustand wird nun auch richtig erkannt, nach dem der screensaver aktiviert oder daktiviert wird.
Feedback ist erwünscht.

no

### DIFF
--- a/include/kodivideochecker/KODIVideoChecker.h
+++ b/include/kodivideochecker/KODIVideoChecker.h
@@ -112,6 +112,9 @@ private:
 
 	/// The JSON-RPC message to check the kodi version
 	QString _getKodiVersion;
+	
+	/// The JSON-RPC message to check the current Playback State
+	const QString _getCurrentPlaybackState; 
 
 	/// The QT TCP Socket with connection to KODI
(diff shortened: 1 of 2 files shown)


brindosch:
Sounds nice! Thank you. Will test and report immediately

brindosch:
It is so amazing, works like a charm.
Even with "stacked" conditions like

grabPause : false
gabScreensaver : true

If you start a movie, click on stop and wait until screensaver starts AND than you stop the screensaver again it switches back to Pause mode as it should!
Also the Method not found is now regarded as Menu. "Method not found" will be always thrown if Kodi is not yet ready to answer or if you use the kodi login screen. So this results in faster overall switching to menu mode after startup for readytouse distributions.

thank you geat work!
Kodichecker is fully working now as it should and grabScreensaver works  like it should (As it never works as it should - at least not 2016)

Issue #23, opened 2016-06-20
cleanup: remove smoothing from colors array


brindosch:
Smoothing should be a own component like blackborder, kodicheck, ...
This is more a small todo note in general :)

redPanther:
... the config part is boring - how about a rewrite and integrated smoothing in  priomuxer - then we have half the way to idle detection, but this a complex task and this shouldn't be done in a hurry

brindosch:
and cause of this i just waited for some suggestions :D

PR #142, opened 2016-07-29
color filter config additions

redPanther:
add ability to disable specific color corrections
update protobuf external
move smoothing out of color section

new options in color section: 
transform_enable
temperature_enable
channelAdjustment_enable

see default config for details

#23

### DIFF
--- a/config/hyperion.config.json.commented
+++ b/config/hyperion.config.json.commented
@@ -54,18 +54,9 @@
 	///            - 'threshold'       The minimum required input value for the channel to be on 
 	///                                (else zero)
 	///            - 'gamma'           The gamma-curve correction factor
-	///
-	/// Next to the list with color transforms there is also a smoothing option.
-	///  * 'smoothing' : Smoothing of the colors in the time-domain with the following tuning 
-	///                  parameters:
-	///            - 'enable'          Enable or disable the smoothing (true/false)
-	///            - 'type'             The type of smoothing algorithm ('linear' or 'none')
-	///            - 'time_ms'          The time constant for smoothing algorithm in milliseconds
-	///            - 'updateFrequency'  The update frequency of the leds in Hz
-	///            - 'updateDelay'      The delay of the output to leds (in periods of smoothing)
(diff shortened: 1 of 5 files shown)


brindosch:
Uh, nice. 
OT: Looks like we need to discuss the whole color adjustments anyway. From HSL(luminance) vs HSV(luminance) over minimumLuminance vs blacklvl and also temperature. :)

brindosch:
If i am right you need to move the smoothing also inside the schema.

redPanther:
thats right ... the temperature thingy annoys me, because it hasn't anything to do with temperature. Its a linear calculated whitepoint.

grrrr ... our test is to good ;-) I had to update the schema files. This is the moment where tests are absolute useful

brindosch:
yeah your tests works pretty good :)

he blacklvl discussion could be done in a way that we may remove the minimumluminance again ang go back to spereated RGB values (At the UI that could be something like "simple/advanced". Your issue here is that the same amount of R/G/B won't work you right? As i see no "real" difference here.

redPanther:
for my taste an easy "blackpoint" and "whitepoint" with linear behaviour is enough for most usecases. Ok it is not perfect, but its easy to code, to use and it's very fast.

For the rgb values we should use 0-255 instead of 0-1. I think this is more understandable and it's easier to use integers then floats.

BTW another topic. I'm sure you know that site: https://github.com/integrations/
I searched for an deployment service. something that will upload after every merged PR to sourceforge.
I found this: https://github.com/integrations/deployhq
It's free for opensource projects.

brindosch:
Hm, okay. deployment from github, but you need binarys to deploy?

Is this one not more our list of interest? https://docs.travis-ci.com/user/deployment

redPanther:
for tags we can use: https://docs.travis-ci.com/user/deployment/releases

brindosch:
Oh, you talked about SF. but i still don't understand who compiles the code for PR deployment :)
Am i miss something?

PR #149, opened 2016-08-03
Update and rename CompileHowto.txt to CompileHowto.md

WoLpH:
Just a documentation update to make it prettier and more readable :)


redPanther:
Thx. One thing is missing. If you do not write the platform param, platform is autodetected. This should fit in most cases

WoLpH:
I mostly reworded and did styling changes but that's might be a good addition as well. Right now I'm still testing on OS X but it actually won't compile for me right now due to spidev.h missing.

The error is understandable but I'll have to see what I can do to fix this :)

brindosch:
Uh, osx! We need your help to get our travis osx build to work, just report or update the travis_install.sh with your brew install commands :D

redPanther:
Write an issue and post compile error. I think we can solve the prob ... yeah and help with trabis would be greay

WoLpH:
Well, this is as far as I can get :)
I have made a few changes to the build script and fixed a small linux specific bit of code.
Additionally, I've added some auto generated files to the ignore list although I'm not a 100% certain if I'm including too much now.

The results from the current run: https://travis-ci.org/WoLpH/hyperion.ng/jobs/150217294
The spidev.h is obviously not available on OS X so we need some way to work around that issue.

redPanther:
You can delete the spi include in ledudpdevice. This was a copy of ledspidevice and have some left overs. This device needs some cleanup. Perhaps @penfold42 can do that?

penfold42:
Oops!

Yep, I'll do it

redPanther:
Fix is merged. Plesse test agsin

PR #151, opened 2016-08-04
sourceOff feature + small json refactoring

redPanther:
add --sourceOff to hyperion-remote - this will select "off" source and set all leds to black
to switch leds back to "on" use --sourceAutoSelect or select a specific source --sourceSelect
-> this is useful for the 24/7 people that complain that their leds blink when nothing is on. (because we send actively black color, leds should stay off)

some tiny refactoring of new json stuff

make schema checker not so strict, do not require values that have defaults (not finished yet)
-> the goal should be that we can start with an empty config and hyperion runs (with default values)
-> this is neccessary for inital setup where the user hasn't configured anything

initialEffect config:
effect is always an array, regardless if it is a color or an effect name
-> this will make transition to qjson a bit easier
made transition to qjson - atm the effect-args are still jsoncpp. To change this effectengine has to be migrated too

webui:
added an example for interactively choosing the input source. Example uses jquery and bootstrap.
 call http://your_hyperion_host:8099/select/index.html
current example needed json server on port 19444
hint: don't forgett to copy latest assets/webconfig files to configured wencnfig dir

yes, initialEffect effect names now as array with one element: ["rainbow"] instead of "rainbow"

@Paulchen-Panther the component state thingy needs some rework. I thing I will make hyperiond class as instance and use this instead of hyperion class. Why? because there we have all the needed objects available. I will do this next week - and I have to say big thanks for new qjson stuff, this is great!

### DIFF
--- a/assets/webconfig/select/index.html
+++ b/assets/webconfig/select/index.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+
+	<title>Hyperion web control</title>
+
+	<!-- bootstrap -->
+	<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
+
+	<!-- Optional theme -->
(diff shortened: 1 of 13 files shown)


brindosch:
Ah, nice yes. Now we need just a good idea how to make this option more accessible. Maybe via json rpc to "set" and "unset" this. Not sure if we find a way for automation. (hm not really)

Great work, moving forward :)

OT: The Hyperion plugin for FHEM is now at their repo, don't ask how i saw this, as i don't use fhem.
https://forum.fhem.de/index.php/topic,55112.0.html

redPanther:
json rpc -> it is accassible via sourceselect

redPanther:
now I added a tiny webui for switching the source, see pr text above.
I also migrate moste initialEffect code to qjson.
that's enough for this PR ... will merge it tomorrow morning

brindosch:
pretty cool! I will test :)
Thank you

redPanther:
> OT: The Hyperion plugin for FHEM is now at their repo, don't ask how i saw this, as i don't use fhem.

I'm pretty sure fhem commuinty will love source select and especially the "off" feature :-)

PR #152, opened 2016-08-05
Fix segmentation fault in usbasp led driver

penfold42:
The usbasp device driver had data type errors in the log calls causing it to segfault with logging enabled.

nope


redPanther:
thx, hope we don't have no more of this stupid bugs ...

PR #155, opened 2016-08-06
Travis osx support

WoLpH:
**1.** Changes

 - Improved build scripts for Travis
 - Fixed small coding error for non-linux platforms

Note that it doesn't fully work yet. The spidev issue is fixed but now it's stalling at v4l. Should I try modifying the cmake files to automatically skip v4l support when it's not available?


redPanther:
v4l shouldn't be compiled ... only dummy in hyperiond. v4l = video for linux ... and it's only for linux as name said ...

In cmakelists for osx: SET ( DEFAULT_V4l2 OFF ) so this should be disabled. can you describe the problem a bit more?

redPanther:
ok I see the problem, some cmake issue .. will look into it later. v4l shouldn't be compiled on osx

WoLpH:
Ok, I've found the problem. Apparently the CMakeLists.txt said V4l2 instead of V4L2 (note the case)

redPanther:
... I updated my message and in the moment I press update you wrote ;-)

WoLpH:
Haha :)

Well, the next problem is the compilation of the main hyperiond source. Another non-linux issue. prctl.h is only available on linux as well

redPanther:
good that we will see such things in future much faster ... did you have a clue how to solve?

WoLpH:
I'm looking into a solution. I don't think there's a cross-platform solution available so I'll have to resort to a few macros I guess. Although Qt might have something. I'm investigating :)

redPanther:
ifdef __APPLE__ is ok, we aren’t so strict in coding rules

WoLpH:
Several solutions... none ideal. Do you have any preference?
http://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits

redPanther:
number 9. I know kqueue from freebsd coding ... this should be the cleanest way. The pipe thingy looks interessting, but a bit more than a hack ;-)

penfold42:
What's the zeroconf problem ?

penfold42:
The code came from https://doc.qt.io/archives/qq/qq23-bonjour.html

And has this on the page;

Aside from the changes to the code above, we also need to link to the Bonjour library on platforms other than Mac OS X. This can be done by adding the following line to the profile.

!mac:LIBS += -ldns_sd

redPanther:
great! big thx!!!!
and compiler warnings showed a bug:
https://travis-ci.org/hyperion-project/hyperion.ng/jobs/150435004#L737
I'll fix it in my next commit. :-)

WoLpH:
Awesome! Great to have it all fixed and merged :)

brindosch:
What a monkey work hehe. Thank you for fixing!

PR #157, opened 2016-08-07
Clone single led color from other led

### LINKED ISSUE #154: Json parser thinks "output" is mandatory
And most but not all led devices require it.

Usbasp doesn't care.

redPanther:
your led stripe longer than needed? some spare leds of same type laying around? Use it with the led cloning feature.
You can make that a led have always the same color as another led.

additional:
- make config json schema more suitable to config (e.g. define enum values for colorOrder)
- make led output not mandatory (fix #154)
- compile warning fixes (detected with clang on osx build)

how to setup:
in your leds array:
e.g. clone color of led with index 0 to leds with index 10 to 12
```
	"leds" :
	[
		{
			"index" : 0,
			"hscan" : { "minimum" : 0.5000, "maximum" : 0.5625 },
			"vscan" : { "minimum" : 0.9200, "maximum" : 1.0000 }
		},
		{
			... leds from index 1 till 9 ...
		},
		{
			"index" : 10,
			"clone" : 0
		},
		{
			"index" : 11,
			"clone" : 0
		},
		{
			"index" : 12,
			"clone" : 0
		},
```

#154


penfold42:
Cool - what does the effect engine see?
The full strip or just the shortened strip ?

redPanther:
good that you ask, I didn't think about. It sees the full stripe, but cloned leds are overwritten afterwards. If we define that clone leds always at the end of a stripe, then its easy to let the effects see the shortend version. If we allow cloned leds in between 'normal' leds, then we have a problem.
Any suggestion how to solve this in a good way?

penfold42:
I can't immediately think of a good way to do it. As long as we document it, I'm ok with it how it is.

I'll think some more overnight...

redPanther:
... now effect sees only the leds definitions without cloned leds. They will be added afterwards.

The performance impact when no clone leds configured is nearly zero. Not tested, but I guess, when lots of clone leds configured, then led cloning can have performance issues. But I don't now where the limit is. All depends on how fast an insert into a std::vector is.

PR #162, opened 2016-08-10
made sure v4l code only initializes when needed and only errors when …

WoLpH:
The code previously always tried to configure the v4l module regardless of whether it was enabled. Now the code remains completely dormant unless requested.

The added advantage is that you don't get a warning that v4l is not compiled unless you've actually tried to enable the module

### DIFF
--- a/src/hyperiond/hyperiond.cpp
+++ b/src/hyperiond/hyperiond.cpp
@@ -536,38 +536,42 @@ void HyperionDaemon::createGrabberOsx(const QJsonObject & grabberConfig)
 void HyperionDaemon::createGrabberV4L2()
 {
 	// construct and start the v4l2 grabber if the configuration is present
-#ifdef ENABLE_V4L2
 	if (_qconfig.contains("grabber-v4l2"))
 	{
 		const QJsonObject & grabberConfig = _qconfig["grabber-v4l2"].toObject();
-			_v4l2Grabber = new V4L2Wrapper(
-						grabberConfig["device"].toString("/dev/video0").toStdString(),
-						grabberConfig["input"].toInt(0),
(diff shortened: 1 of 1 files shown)


redPanther:
not accepted. when build with ENABLE_V4L2 then in every case v4l object has to be created, only starting should depend on enable flag. We need that for start/stop/configure v4l at runtime.
suggestion. leave code in ENABLE_V4L2 mostly as it was, pull out grabberConfig var and in `#else` branch add a && grabberConfig["enable"].toBool(true)

redPanther:
... or just like that:
```
#else
	const QJsonObject & grabberConfig = _qconfig["grabber-v4l2"].toObject();
	ErrorIf(_qconfig.contains("grabber-v4l2") && grabberConfig["enable"].toBool(true),
	        _log, "The v4l2 grabber can not be instantiated, because it has been left out from the build");
#endif
```
best would be if we can remove ifdefs ... ifdefs have lot of disadvantages, but in this case there are too helpful

WoLpH:
Hmm, that makes sense but I believe the previous code was incorrect in that case. Unless the configuration option was defined it wouldn't be instantiated.

Removing the ifdefs entirely would be impossible since the v4l2 includes are simply unavailable but we might be able to reduce it slightly further. I'll investigate :)

redPanther:
... just do the small solution I suggested.
The code was from time where v4l hadn't a enable flag and on linux systems we usually compile with v4l. That's why nobody had cared about that lines of code.

We can remove ifdefs if we have dynamic linking at runtime. (with dlopen) - or we have dummy libs when original libs are absent.
But this topic is part of another issue. This comes later ...

WoLpH:
Fair enough :)

Let's leave it for a later time. The notification was simply bugging me ;)

PR #163, opened 2016-08-10
display the currently selected color in color

WoLpH:
Just a small change for the webinterface. I thought it would be nice if the input field actually shows the selected color. To illustrate:

image


redPanther:
cool, thx!

Issue #148, opened 2016-08-03
Runtime error if color number is out of range (JSONConnection)


b1rdhous3:
Log:
```
[HYPERIOND JSONSERVER] <DEBUG> <JsonServer.cpp:64:closedConnection()> Connection closed
terminate called after throwing an instance of 'std::runtime_error'
  what():  unsigned integer out of signed int range
Aborted
```

Command:
`{"command":"color","priority":1,"color":[142333333333333,331,1],"duration":3}`

redPanther:
.. then put no big numbers in it ;-) 
we are in process to change our json parser. We will fix this issue in that process.

b1rdhous3:
But I am always curious what happens if I do ;)
Okay, sounds good.

brindosch:
Looks like no one before tried to feed Hyperion with "out of scope" values :)
No problem, it is good to catch and note everything. In best case now :)

redPanther:
I'm not happy with those plenty uncaught exceptions - thats another todo. But it can be more worse - we have exceptions in code parts where exceptions not allowed, e.g. in v4l grabber. I'm happy that we rework so much, but it feels everyday we find new stuff that can be made better ....

redPanther:
... for exactly that tests we need automated unittests, that check all corner cases of the code ...

>  , but it feels everyday we find new stuff that can be made better ....

PR #164, opened 2016-08-10
fix uncatched exception in json message handling

redPanther:
fix uncatched exception in json message handling

#148 

### DIFF
--- a/libsrc/jsonserver/JsonClientConnection.cpp
+++ b/libsrc/jsonserver/JsonClientConnection.cpp
@@ -216,55 +216,64 @@ void JsonClientConnection::handleMessage(const std::string &messageString)
 {
 	Json::Reader reader;
 	Json::Value message;
-	if (!reader.parse(messageString, message, false))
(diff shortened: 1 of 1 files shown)


redPanther:
@WoLpH please test

WoLpH:
The change looks good but I think it would be better to remove this block entirely: (https://github.com/hyperion-project/hyperion.ng/pull/164/files?w=1#diff-517591e3d8a9057c992935a4b65e68fbR222)
```
if (!reader.parse(messageString, message, false))
{
	sendErrorReply("Error while parsing json: " + reader.getFormattedErrorMessages());
	return;
}
```

The `checkJson` function does a `reader.parse` internally so it's a bit redundant now.

redPanther:
both blocks needed. first one is for the message, second one is for the schema file

WoLpH:
Ah, you're right. I didn't look thoroughly enough. In that case simply a 👍  from me :)

PR #167, opened 2016-08-10
Added some Travis caching

WoLpH:
It doesn't really do all that much but it does speed up Travis builds slightly by enabling ccache and the brew download cache for the OS X build.

It's still slow, but less so :)


redPanther:
thats cool thx!

PR #168, opened 2016-08-11
v4l grabber wont crash hyperiond anymore

redPanther:
v4l errors won't lead to crash hyperiond (exceptions are catched)
v4l auto device mode with better log messages
createV4l in hyperiond.cpp refactored, now v4l is always available as long as it is compiled in

to test the exception thing, start hyperiond with v4l, then pull usb plug of grabber, now v4l should be switched off. Currently it is not possible to reenable it via hyperion-remote/json. This will be possible in future ...

additional:
fix compile of dispmanx (bug was introduced with latest grabber rework)


brindosch:
This is great :)

PR #170, opened 2016-08-12
Initial WebUI design and structure with JSON sample function

b1rdhous3:

Initial WebUI design and structure with JSON sample function


brindosch:
That's hot!

b1rdhous3:
Your button has arrived! ;)

brindosch:
Wuhu :)
Thank you

redPanther:
thats great!!! While fixing the kodi beast, I checked out your branch, set the path in my config.json and it worked out of the box. I can set the color ... yeah.

BTW current remote control ui is only a place holder, If we/you want to replace that thingy ... I would love it.

brindosch:
I said he could move all examples to subfolders, they are helpful for now. (At least i think).
We could remove them later.

I like the already working translation :D

redPanther:
... yeah that's good.

b1rdhous3:
I will take the remote example to build something like that in the WebUI. I really like that color wheel

redPanther:
I like that one:
 http://mathias-biedert.de/wp-content/uploads/2015/11/webapp.png
https://github.com/poljvd/hyperion-webapp

I use it also on my phone and on my desktop.

b1rdhous3:
Do we really need the option to set the prio so slightly? 
I think it would be confusing to change it that way. It would end in hundreds of prios. 
Maybe a default value?

But all in all i like that

redPanther:
the prio thing is not needed. One value, set in config (the dashboard, not in the remote itself) and thats it ...

redPanther:
this color picker is used
https://github.com/mattfarina/farbtastic

brindosch:
Yes, 2014 and that way it looks like ;)
Aww you are right, clear all kills our background-effect. This is bad! This is the only way to get rid of prios you set "somewhere".

penfold42:
Maybe we need a Clearall-mine option ?

Now that we can track the source of the priorities could we clear all that are my source ?

redPanther:
If you have an app (web,phone), there you should set the prio in config. then we can use normal clear, to clear my channel. If the remote app let the user set multiple channels, then the app has to keep track of that.

we can switch "off" the leds by selecting the "off" source.

redPanther:
@b1rdhous3  should I merge, or should I wait for more commits.

b1rdhous3:
If you like to merge you can do it. 
I think this state is a good start for the webinterface.

PR #171, opened 2016-08-12
Kodi initial fix

redPanther:
another fix for kodi initial state. Now it should work when on startup video is playing a or menu is displayed.

### DIFF
--- a/include/kodivideochecker/KODIVideoChecker.h
+++ b/include/kodivideochecker/KODIVideoChecker.h
@@ -166,5 +166,8 @@ private:
 	/// flag indicating state
 	bool _active;
 
+	/// flag indicates if playbackState is valid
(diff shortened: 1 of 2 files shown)


redPanther:
.. we should drop to ask kodi what's the state is. We should make a kodi plugin that has options to set what should happen in which state and then send this to hyperion via hyperion json rpc.
-> then we are not depending on complex kodi json protocol and never think about kodi version in hyperion. (we jst have to maintain our kodi plugins.)

Perhaps our current plugin can be extended with that function ...

penfold42:
In some ways I prefer the current approach even tho the code is pretty horrible.

It simplifies the install and config process for end users and you have have multiple Hyperion instances easily getting state rather than having the maintain install and configure another component

PR #177, opened 2016-08-14
Initial Hue Bridge support & Small changes

b1rdhous3:

Added Hue Bridge support in WebUI (see GIF)
hue


brindosch:
Nice, ah one question. isn't this a "Wizard" for the Wizard category? (Or tools)
Just an idea :)

b1rdhous3:
Since we don't know yet how to do the device setup (and this is a part of the hue setup) I think it is good for now. 
We should talk about that topic soon.

PR #179, opened 2016-08-15
add command for getting schema via json api

redPanther:
- update coding style
- add command for getting schema via json api
- json api: merge config commands into one single "config" command with subcommands
```
{"command":"config","subcommand":"getconfig"}
{"command":"config","subcommand":"getschema"}
{"command":"config","subcommand":"setconfig","config":{"logger":{"level":"warn"}},"create":false}
```


b1rdhous3:
Can't wait to test! :)

redPanther:
BTW you can get the configured json port via jquery ajax call from url:
"http://"+document.location.hostname+":8099/cgi/cfg_jsonserver"

result:
:19444

Paulchen-Panther:
Great. Super. +1

b1rdhous3:
@redPanther I think we need a way to send the tanslation token and default language key for our translation with the schema.

redPanther:
Default language is the browser language.
You can add 'title' in json schema and use it for translation token. 
We can save language as cookie. If no cookie, show language selector with browser lang preselected.

brindosch:
This may be possible for language but i am not sure if this is the long run in general (other options/flags that you don't want in config .json)
So maybe another settings.json to read from for system settings or a database like you mentioned for better effects handling/configuration

b1rdhous3:
Our translation lib needs to tags in a html element like <a> <span> etc:
data-lang-token and lang. 
But I think we also could set these two with JS tag manipulation after the JSON Forms were generated

redPanther:
In config.json we can save webui related stuff in webconfig section.

PR #185, opened 2016-08-18
Example JSON Form with General settings

b1rdhous3:
I have created a new Category "General" with several Settings for
Hyperion.
For now the Submit Button only console.log's the values.

Known Bug: For some reason it seems to do not load the bootstrap
correctly or maybe the Lib isn't compatible with the newest version of
bootstrap.


redPanther:
yeah!

PR #186, opened 2016-08-18
Update package creation

brindosch:
 - Append platform in filename which determine the origin. Now you could upload all package at the same time without confusion
 - bins now at correct location inside packages
 - update .deb with qt5 dependencies
 - modifiy slighty postinst
 - update crosscompile with new deps and latest hints :)

There is still enough to do. I will test these packages soon for further improvements
@redPanther Is there a reason why you call install_file with a .new append? Counts this also for services?
Hint: It is possible i misuse here something ;)

With our webui there comes more questions along the road anyway :)


redPanther:
This leaves user modified config files in place and don't overwrite them with the new file. Instead new file would be renamed to *.new
I don't like that packages overwrite my personal config ;-) -> read the defined shell function "install_file" in that file. Then you can understand what that thingy does.

For the webui this must be think over. I think about a config overlay. A posibility: in /etc/hyperion is always the initial config and in /var/..../hyperion there is the file created by the webui.
If we do so, we won't have trouble with permissions, because writing to var is ok with none root access. But we should delay this and focus on the next steps ....
(e.g. mysql writes there DBs to /var/lib/mysql)

PR #188, opened 2016-08-19
Translationsupport for JSON Forms & Bootstrap Switches

b1rdhous3:
Bug when German is selected and you click on the Button for General Page again. 

Someone a idea for fix?


b1rdhous3:
I just added Bootstrap switches for all checkboxes in this PR too.

PR #209, opened 2016-08-30
Web: Centralize websocket, add json port changer and led schemas

brindosch:
 - centralize websocket request
 - Adds a simple json port changer, in case of failure a modal opens which tells you that.
 - hopefully fixing the hue modal
 - Adds some json led device schemas (far from done :) )
 - Dash: add current led device info and host:port info


b1rdhous3:
Wow! That are many changes. nice!
But don't we want to get the schemas from hyperiond via JSON RPC? Or am i getting something wrong here?

brindosch:
Oh, the "Web:" was misleading. It is: libsrc/leddevice/schemas/schema-apa102.json

redPanther:
Schema thingy os absoluty ok so.

redPanther:
Small schema thing:
All devices support param ledCount. 
Can you add new param default for every property? Value must be taken from code.

Example:
"default" : "/dev/ttyACM0",

Later on cpp code will take default values from schema

b1rdhous3:
Ah I see. Okay :)

redPanther:
... and you can get the hyperion json port from webserver via ajax. No need for hardcodef value. Will give example code later ... writing this via phone is to hard

b1rdhous3:
Yeah I know how. You have told me before. I can take care of that too :)

redPanther:
Oh I see defaults are already there ... sorry overlooked it ....

brindosch:
default: is partly implemented (blank fields are unknown or unsure)
ajax/js: still new to all this. structrue is horrible and far from clean :)
I want to improve it further.

We may change the forms.js again, as the current one is scruffy since 2014 and has no support for boostrap 3 (which is a lesser problem). Also the structe is very strict and it has pretty much features we never need.
I like the approach here more: https://github.com/jdorn/json-editor
As everything is more splitted it is much easier to extend it to fit our needs. There are some PRs pending like better translation support and other stuff.

Translation: Too much issues (nesting html elements) with the current implementation and lacking of support to include vars (Which is sometimes better for a good looking feedback).
--> We need another system here

PR #211, opened 2016-08-30
Travisci: add tag and cron support

brindosch:
So after some support contact and doc reading
 - We will get tomorrow a enabled cron for daily builds

deploy is still open, will clear it up too. If i get what i want this will end up in a nice auto building and deployment function.
@redPanther sorry, i need further help regarding pi crosscompile.

[skip ci] <- skips ci for PRs if you don't want it


redPanther:
yeah! great!

PR #213, opened 2016-08-31
Added debug logging to ProviderSpi.cpp on writes.

penfold42:
If you have more leds configured than the SPI buffer can support you now get this:
[HYPERIOND LedDevice] <DEBUG> <ProviderSpi.cpp:92:writeBytes()> SPI failed to write. errno: 90, Message too long

nope


penfold42:
Don't merge this yet

WoLpH:
I think this should actually result in a fatal error as it won't do anything without any apparent reason.

penfold42:
It's now error - it can't be fatal to the point of killing hyperiond - we need to be able to reconfigure things on the fly

PR #214, opened 2016-08-31
fix amlogic for newer kernels

redPanther:
fix amlogic grabber for aarch64 architecture (e.g. wetek hub, wetek play2 / S905 cpu)
crosscompile for wetek core/play should work as expected. For newer boxes we haven’t crosscompile, because it is based on 32bit raspbian for rpi .... user for such boxes have to compile hyperion.
Perhaps other kernel config is needed, see referenced issue.

@stefansaraev fyi

no

[#550](https://github.com/hyperion-project/hyperion/issues/550)

crosscompile stay on 32bit aml with kernel 3.10


stefansaraev:
thanks. just a side note:

amlogic 3.10/3.14 kernels also have amstream.h but that's a private header that should not be used, users with s905 and newer boxes should have "libamcodec" headers installed. if they dont have the headers in their /usr/include (or cross toolchain) - the resulting binaries will silently fail at runtime. distros who do their own packaging should take care. //cc @lrusak add libamcodec as a dependency - this will work with older aml boxes including wetek play and core.

also, regarding the required kernel changes, I didn't have to do any. hyperion (I am using "slightly" modified sources based on old hyperion) just worked ootb for me on wetek play2 after I fixed AMSTREAM_IOC_GET_VIDEO_DISABLE define. sure, amvideocap fails when playing 4k material (colors look inverted, but they are not really inverted, it's just that amvideocap does not work yet with 4k, it produces garbage, I'll report the issue to codesnake)

redPanther:
Perhaps it is better to not use include, just switch the AMSTREAM_IOC_GET_VIDEO_DISABLE cia cmake. "if cpu is amlogic and arch is 64bit -> then "int" else "unsigned long" for AMSTREAM_IOC_GET_VIDEO_DISABLE

perhaps this is better?

stefansaraev:
> if cpu is amlogic and arch is 64bit

nah, using the header is the best way imo. there is no guarantee that AMSTREAM_ defs will never change in the future. 

c/p from external headers that are subject to change is always a bad thing ;)

redPanther:
> c/p from external headers that are subject to change is always a bad thing ;)
yes of course ... ok then we take current version as a start. 
BTW we determine cpu during cmake stage via /proc/cpuinfo.
I will disable integrated protobuf for amlogic 64bit cpu, so we didn't run into problems for now.
If you find the time you can create a bug report with the compile error and I will fix it.

redPanther:
now compiling on wetek machine is:
cmake -DCMAKE_INSTALL_PREFIX=/usr .. && make -j4

this will select platform "amlogic64" for newer boxes.. This enables aml grabber and use system protobuffer instead of intergated

you can also install hyperion with: make install (or on debian call make packages for a deb package)

PR #218, opened 2016-09-02
Add more schemas + webui integration

brindosch:
 - integrate new schemas and add to to ui
 - there are sequence issues, you see this now better than before. Solution necessary as everything changes depending on browser. A sequence is ususally just given in an array, this could be a part of the issue.
 - not all schemas are done
 - v4l2 throws an error, no idea why.


redPanther:
we shoud add another flag in schema named "advanced" to mark settings that are to "nerdy"

atm I transform all websocket stuff to event based behaviour. I will merge this PR and make needed changes (otherwise I get into conflict hell ;-) )

PR #223, opened 2016-09-04
Webui: view led configuration + lights

redPanther:
add view of ledconfig and live led colors (update every 2 seconds)


b1rdhous3:
Nice!
Wouldn't it be good to have the option to change the refreshrate via config? For example in webconfig schema part. So you can change it directly in the webUI

redPanther:
After thinking about led color update I realized that current way of getting the data (request-response) is not the good modern way to do.
I will change this that we send a request to start led data transfer (with interval in ms) and then hyperion pushes constantly leddata messages to webui until stop request recieved or websocket closed. Same thing I will do with the serverinfo.

brindosch:
one word: amazing!
Could you change height to 450px which will then fit 16:9 ;)

redPanther:
... change to 16:9 and prepare for realtime leds + other enhancements

redPanther:
Rework of push messages for serverinfo and led streaming comes in next pr

PR #224, opened 2016-09-05
webui: enhance led config

redPanther:
- split content and js
- tune leds config: now with tabs and custom config edit (save not working yet)
- add connection lost detection


brindosch:
impressive!
well, the editor is a little bit laggy if you scroll down the first time (afterwards it is something like "cached?")

redPanther:
call for a good connection lost page. Be aware that all things must be loaded on startup, because when connection lost, you can't get any data from server.

automatic check if server is reachable and then redirect to server would be nice
file: content/connection_lost.html

brindosch:
Well, adding a picture with Hyperion and a broken connection to browser logos would be easier than fiddling around with font awesome and would make the message more clear.

Workaround: add img to index.html and set it to display:none, now you could access the picture via css (background-image) even if websever is down. To change picture sizes depending on screen size use @media property

redPanther:
... see #226

brindosch:
It is still mandatory to load the picture once

redPanther:
of course ... there is a hidden div where we put all connection los data

PR #236, opened 2016-09-11
extend schema for leddevices

redPanther:
extend schema for led devices a bit. still lot's of work here ...

### DIFF
--- a/.travis/travis_deploy.sh
+++ b/.travis/travis_deploy.sh
@@ -4,7 +4,7 @@
 sf_upload()
 {
 	/usr/bin/expect <<-EOD
-	spawn scp $1 $2 hyperionsf37@frs.sourceforge.net:/home/frs/project/hyperion-project/$3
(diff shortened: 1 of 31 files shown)


brindosch:
I vote for remove this format:checkbox option. I "hardcoded" it at my branch: https://github.com/brindosch/hyperion.ngBeta/commit/2ac86d040b00d0a609e5971413c9672c82d7b8ae

redPanther:
oh great! I don't understand why jsoneditor has this bad default behaviour for booleans ...

redPanther:
BTW if you modify jdoneditor. Would be great if you can alter behaviour of headlines. If headline has empty title, the headline (h3) is not shown. currently we add a title: " " to supress headline, but it add more space then needed

brindosch:
This is the general fallback, if no title attribute is available, it falls back to object description. In the meantime i found the fallback, now it is empty, but the h3 is still rendered. need to catch the non existing title attribute to prevent this.

redPanther:
instead of throwing out the h3, set css("display","none") for hiding them. Then dom is not altered, but we get rid of unneccesary space

PR #237, opened 2016-09-12
effects included in hyperiond  binary as qtResource

redPanther:
effects now included in hyperiond binary via qt resource (like the json schema files)
-> no need to install effects in system
-> no need to set effects path (unless you want to make your own)

when loading effects from multiple path, then only last effect with same name is loaded. (e.g. you have the rainbow in multiple locations: in hyperiond, in /usr/share/hyperion/effects, only the effect last in list is loaded)

want to modify only a param in json file? Then:
- add a path to effects to location where you want to place the json effect definition
- copy json file from hyperion source
- modify params you want
- set a : before script name. e,g, ":fade.py" instead of "fade.py"

want to hide a effect, that it doesn't appear e.g. in webui/json/phone app?
in hyperion.config.json:
```
	"effects" : 
	{
		"paths" : [ "efx" ],
		"disable" : ["Rainbow swirl", "X-Mas"]
	},
```
.. the disable param. It is an array with effect names that should be disabled

 finally solve open file handle during effect is playing. Now script is read before, then file closed and then t is run by python.

additional:
webui: add write support for led controller setup

optionally, look above


penfold42:
Cool idea - I've been caught a few times with a missing effects dir !

redPanther:
later on we can do this with the webui. Then webui always find his files and nobody can play around with webui.

Paulchen-Panther:
Also diesen schritt finde ich persönlich Unfug. Als erstes versuchen wir die Dateigröße von Hyperion runter zu bekommen und alle grabber auszulagern - danach wurden doch alle grabber wieder intern, und jetzt dieses hier.
Dachte echt wir steuern auf ein Plugin, addon oder wie man es sonst nennen dürfte System zu aber anscheinend nicht. Schade

tpmodding:
i agree with paulchen, we could include 3-4 standard effects, but not more. the folder option should stay.

redPanther:
of course folder option must stay, because I want to overlay standard effect params!

Issue #122, opened 2016-07-16
Dynamic link against bcm_host and x11


redPanther:
Bind (dlopen) against bcm and x11 libs only when dispmanx/x11 is used. 

This enables us to e.g. to launch raspi build on other arm linux, where bcm libs not available.
Or compile with x11, but hyperion is runable on system without x11 ..

Then we don't have to remove integrated grabbers and use only proto grabbers.

brindosch:
Hm, can't follow
> Or compile with x11, but hyperion is runable on system without x11 ..

eg Pi, but you can't enable dispmanx without Pi sources so it remains the same (-> Proto)?
Under the hood you could use a hyperion-x11 enabled internal build with a dispmanx proto implementation?

if platform grabber not inside the build - call the external one?

redPanther:
ok try to make a simpel example.
compile hyperiond on raspi with x11 and dispmanx and amlogic.
compilation has more or less 3 stages. configure, compile and link.
during link phase do not link against bcm (dispmanx) and x11 libs (currently we do).

If linked against e.g. x11 then it is neccessary that x11 libs are in place when start hyperiond. If libs not there hyperiond will not start and throw error "missing x11 libs"

now the new part: if not linked on binary creation, those libs arn't needed to run hyperiond, as long as the code part for x11 grabber is not executed. when config says enable x11, then we load x11 libs during  runtime. if loading fails, throw error: x11 not usable .. same for dispmanx.

we can use dispmanx enable binary on systems where dispmanx is not available.

brindosch:
Ah now i understand :)
Thank you. 

OT: For sure a "plugin system" is a nice thing. Like HDMI-CEC stream reading and assign actions, or IR signals.

redPanther:
After investigating a bit I had some additions:

currently all grabbers are made as static libs. This is generally good, because we need the grabbers as lib for "load on demand".
Todo: the graber libs have to build as shared lib. Then we can load the grabber (and indirectly x11 and bcm libs), if we need it - if loading of grabber fails, then x11/dispmanx is not available and those grabbers are deactivated.

tpmodding:
@Paulchen-Panther ist der durch?

Paulchen-Panther:
Ich glaube schon

PR #239, opened 2016-09-13
make shared libs work again

redPanther:
resolve cyclic dependencies between hyperion modules. Now we can compile as shared libs (again).
This is needed preparation for dynamicly load (dlopen) grabbers or other components.

```
cd build
cmake -DBUILD_SHARED_LIBS=ON ..
make -j$(nproc)
export LD_LIBRARY_PATH="lib:$LD_LIBRARY_PATH"
bin/hyperiond --debug path_to_my_config.json
```

generated libs (for platform x86-dev):
```
libamlogic-grabber.so
libblackborder.so
libboblightserver.so
libbonjour.so
libcommandline.so
libeffectengine.so
libframebuffer-grabber.so
libhidapi-libusb.so
libhyperion.so
libhyperion-utils.so
libjsoncpp.so
libjsonserver.so
libkodivideochecker.so
libleddevice.so
libprotobuf-lite.so
libprotobuf.so
libprotoc_lib.so
libprotoserver.so
libtinkerforge.so
libudplistener.so
libv4l2-grabber.so
libwebconfig.so
libws281x.so
libx11-grabber.so
```

somehow related to #122 

### DIFF
--- a/libsrc/hyperion/CMakeLists.txt
+++ b/libsrc/hyperion/CMakeLists.txt
@@ -58,10 +58,8 @@ add_library(hyperion
 )
 
 target_link_libraries(hyperion
-	kodivideochecker
 	blackborder
 	hyperion-utils
 	leddevice
-	effectengine
 	${QT_LIBRARIES}
 )
--- a/src/hyperion-aml/CMakeLists.txt
+++ b/src/hyperion-aml/CMakeLists.txt
@@ -30,7 +30,8 @@ add_executable(${PROJECT_NAME}
 )
 
 target_link_libraries(${PROJECT_NAME}
-    commandline
+	effectengine
+	commandline
 	blackborder
 	hyperion-utils
 	protoserver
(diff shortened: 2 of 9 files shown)


redPanther:
@Paulchen-Panther  we will have some sort of plugin system, but currently we are busy with basics. E.g. my big grabber refactorings (making common base classes, unify interfaces ...) are steps to more flexibility and a way to make a good plugin system.
Or the switch to qt components ....

The first idea for this topic was to use only standalone grabbers, but this isn't good because
- this solution can't be used for all components
- using proto as ipc is an unnecessary overhead, when used local, 
- dealing with child processes is no fun
- adds complexity instead of making it simpler

We will have to sorts of plugin types: python scripts (we already have, the effects, and hopefully soon webserver backend scripts) and runtime loadable cpp libraries with well defined api. But i fear this have to wait a while. This task is to complex as we should do that to hasty.

Paulchen-Panther:
Its Ok. Sorry for misunderstanding.

PR #249, opened 2016-09-21
extend effect engine with qt image effects

redPanther:
extend effect engine for better image drawing effects:
- hyperion calculates led layout grid size from led configuration
-> this is needed to create optimzed image dimensions for image effects
- now every effect creates a image (qimage) buffer with optimized size (for optional use of course)
- this image buffer can be modified with qpainter commands and when drawing is done, image buffer can be send to the leds
- currently QCanonicalGradient and QRadialGradient are implemented, but all other commands are possible now, just some monkey work (qpainter functions are e.g. drawing lines, circles, rects, text, gradients, ...)

The rainbow effect is now image based. The advantage is, that it also look good on other matrix or more strange layouts. 

here main code from rainbow-swirl.py
```
	hyperion.imageCanonicalGradient(0,0,
		hyperion.imageWidth,hyperion.imageHeight,
		int(round(hyperion.imageWidth)*centerX),int(round(float(hyperion.imageHeight)*centerY)),
		angle,rainbowColors
	)

	hyperion.imageShow()
```
the **imageCanonicalGradient** func takes several arguments: 
"rect of drawing area - startX, startY, width, height" "center point - cx, cy" "gradient angle" "list with gradient colors"*

*
needs 4 values per color: position (0-255), r,g,b
more info here:
http://doc.qt.io/qt-5/qconicalgradient.html  (stop color)
http://doc.qt.io/qt-5/qradialgradient.html

also other gradients are possible, but not yet implemented:
http://doc.qt.io/qt-5/qlineargradient.html

after drawing on image buffer, the image buffer must be send to the leds with
**imageShow**

hint: play with the center point in rainbow-swirl json file. 0.5 means center, values are valid from 0 to 1

bildschirmfoto5

additional:
fix some js errors

yes - for better js compatibility "grabber-v4l2" is renamed to "grabberV4L2"


penfold42:
Good work.

How does it behave with non rectangular led definitions ?

I have a 60 led ring and also a 61 led set of concentric circles (24, 16, 12, 8, 1)

redPanther:
the image is projected onto your led layout, the corners are of course truncated. But a good effect should be working (= looking good) on different layouts. E.g. "knight rider" works on every layout - former rainbow swirl worked on all frame like layouts (regardless which form (rect/circle/..) ), but not on matrix like layouts.
Effect developers should test effects with diferent layouts (e.g. with hypersim, to save hardware).
We should collect some sort of different led layouts in config file. e.g. different matrix types, your circle ...

penfold42:
Good idea - a sample/contributed directory with no warranty.
I've got some spreadsheets I used to generate configure too I'd be happy to share

redPanther:
... a subfolder in assets would be nice. I also have a python script to generate my special led layout. Would be nice if you can take over this job and commit your stuff ... :-)

brindosch:
How big are these "empty" edges? I tested this now on my usual setup and i can't see them (which is good).

PR #251, opened 2016-09-21
add ability for default values in providerudp

redPanther:
add ability for default values in providerudp
extend/fix udp ledevice schema files
add default port for tpm2net according http://www.ledstyles.de/ftopic18969.html
move switchOff into baseclass

no

### DIFF
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -47,3 +47,15 @@ MyClass::MyClass()
 {
 }
 ```
+
+- pointer declaration
+
+```
+bad:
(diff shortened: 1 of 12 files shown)


penfold42:
Looks good

What happens if you try to send to port 0 ?
I know listening on port 0 is ok - you get a random port - but I've never tried sending to port 0

redPanther:
Never thougt about doing that, I hope users don't try that.

brindosch:
if so, raise the min value :)

PR #256, opened 2016-09-25
Dmx512 Rs232 support

penfold42:
Add DMX512/Rs232 support

```
        "device" :
        {
                "name"       : "DefaultHyperionConfig",
                "type"       : "dmx",
                "output"        : "/dev/ttyUSB0",
                "rate"  : 250000,
                "colorOrder" : "grb"
        },
```

### DIFF
--- a/libsrc/leddevice/CMakeLists.txt
+++ b/libsrc/leddevice/CMakeLists.txt
@@ -33,9 +33,10 @@ SET(Leddevice_HEADERS
 	${CURRENT_SOURCE_DIR}/LedDevicePaintpack.h
 	${CURRENT_SOURCE_DIR}/LedDevicePiBlaster.h
 	${CURRENT_SOURCE_DIR}/LedDeviceSedu.h
+	${CURRENT_SOURCE_DIR}/LedDeviceDMX.h
 	${CURRENT_SOURCE_DIR}/LedDeviceFile.h
 	${CURRENT_SOURCE_DIR}/LedDeviceUdpRaw.h
-    ${CURRENT_SOURCE_DIR}/LedDeviceUdpH801.h
+	${CURRENT_SOURCE_DIR}/LedDeviceUdpH801.h
 	${CURRENT_SOURCE_DIR}/LedDeviceUdpE131.h
 	${CURRENT_SOURCE_DIR}/ProviderUdp.h
 	${CURRENT_SOURCE_DIR}/LedDeviceHyperionUsbasp.h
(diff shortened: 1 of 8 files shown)


redPanther:
Great I have all hardware to test, but My problem os, how to setup dmx address, rgb to fixture capabilities ...

penfold42:
Excellent! I can't test beyond seeing serial data leave my ftdi adaptor.

It's dumb - no led / slot mapping - Hyperion led 0 is sent in slots 1,2 and 3 (r,g,b), Hyperion led 1 in slots 4,5,6 and so on

redPanther:
I have a Mc Crypt LED-spot. This is DMX spec:
```
DMX channel | Function
1           | Intensity red 0-100% (0-255)
2           | Intensity green 0-100% (0-255)
3           | Intensity blue 0-100% (0-255)
4           | DMX value
            | 0-7     Light off
            | 8-190   Dimmer
            | 191-200 Sound-to-light control
            | 201-247 Flash effect
            | 248-255 Light on
```

When color to set is the same color as it is currently set, but ligther/darker, then the use of the dimmer is prefered, because it is faster and more smooth instead of setting the rgb channels individually. It is like the white algorithm for rgbw leds.

Most DMX stuff I know has at least 4 channels R-G-B-Specials (of course this is for colored light).
You should add at least a channel gap and a start/end address.
I mean something like start on dmx address 10, then next channel rgb (10,11,12) the gap channel (e.g. 2 -> 13,14) and then start again with rgb (15,16,17) .... until end address is arrived.
Perhaps instead of end address a "fixtureCount" (= how much led spots you have) is better.

setting an off value can be a nice thing to. this should be notated in "channel:value" e.g. 4:0 in my case. in case of "off" all spots (=> fixture count) should set to off. example for 4 spots: send to channel 4,8,12,16 the value 0. (when start address is 1)

In the long run, better fixture config would be the best. (but some kind of complex) hey and then hyperion can go to stage :-)

In future I'm pretty sure we will have music visualization, imagine using that with dmx fixtures for your party. .... endless thoughts whats possible ...

penfold42:
I'll ate the start address and the increment.

The e1.11 DMX512 spec makes no statement about how the 512 bytes are interpreted - they are just 512 "dimmer" channels that may or may not have specific significance.

I think we'll need a big case statement to support various "DMX-light type"

redPanther:
Thx. This schould be enough to drive some fixtures.
For serious dmx support we should read fixture specs from qlc:

https://github.com/mcallegari/qlcplus/blob/master/resources/fixtures/

There is a neat editor to create this files.
In config there should be an array of fixture names that correspond to name of spec. The array idx of fixture correspond to led id.

Ok enough with ideas :-)

Have a nice day

penfold42:
I've added support for yours!

Well, I would have if it didn't segfault

penfold42:
Segfault fixed

penfold42:
I'm guessing this is yours :
http://www.conrad.com/ce/en/product/591490

I'd love to see if my code works in real life !

redPanther:
it basicly works, but timing is not accurate enough and the refresh rate is to low. In general we should put the leddevice in an own thread, so device won't disturb the main thread and vice versa.

With qlc I get it working failry ok. Problem is, that the pc has to generate a timing correct signal and this is mostly not acurate. (same thing with driving ws2812 on raspi with normal gpio)

Making a dmx device with usb rs232 + rs485 converter is known to make problems. Better solution is that a mcu makes the dmx signal and you transfer the light data to the mcu.
I build a adalight2dmx sketch, but currently it is also not working well enough (but not investigated whats the issue)

Nevertheless the device controlled by qlc works much better as our dmx driver. I alredy planed to add more threading to hyperion. I will add for leddevices soon and then we will see how it performs.

redPanther:
thats mine: https://www.conrad.de/de/led-pinspot-dmx-led-pinspot-anzahl-leds-36-x-591241.html

redPanther:
I cheated a bit and used the smoothing to get better output refresh rate. This was ok to maintain a steady clock signal, but sadly not accurate enough to transfer data.

penfold42:
That wa my biggest worry.

the break and make after break is currently done with nanosleep - there could be framing errors here where a picky device ignores the update of the timing is out.

If we move to a separate thread we should look to move the continuous output function to that thread.

PR #259, opened 2016-09-30
fix input selection wont work when color is set

redPanther:
... fix input selection wont work when color is set

test:
- start hyperion
- open webui -> remote controll
- set a color
- try to set a specific input source
-> current: led color stays always
-> fixed: led colors are set according selected input source

### DIFF
--- a/libsrc/hyperion/Hyperion.cpp
+++ b/libsrc/hyperion/Hyperion.cpp
@@ -828,7 +828,7 @@ void Hyperion::setColors(int priority, const std::vector<ColorRgb>& ledColors, c
 		_muxer.setInput(priority, ledColors);
 	}
 
-	if (priority == _muxer.getCurrentPriority())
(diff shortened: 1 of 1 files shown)


brindosch:
cool!
Another question:
If you set a source and this source "dies" (like clear or no longer input data via proto). It keeps this source as active, how should it behave?
- Keep it as is (as the user is responsible to handle that)
- fall back to "auto" if selected source dies

Thanks for your input :)

redPanther:
currently: if an active input "dies" it will be removed from priority list and input-auto-select is reenabled.

brindosch:
Strange, it works. Wrong memory, it was just the issue with colors.
Good!

redPanther:
when color set it didn't update the current input source variable. that's why the whole thing get messed.
solution: when in "no auto" mode and color is set, update all stuff

PR #260, opened 2016-09-30
light clock

redPanther:
- extend image effects
- add a basic light clock - this is not the final state, but a start and an invitation for ideas

... this effect respects matrix layout and orientation.


brindosch:
cool!
Q: Is alpha good or bad? As alpha ist not really a difference to black? (In our led case)

redPanther:
Alpha is important. Look I overlay 3 gradients, if we dont have alpha channel we only would see last gradient. And it is not neccessaeily black. It depends on how is the image backhround (solidfill)

redPanther:
dscn0224

PR #263, opened 2016-10-07
Leddevice refactoring the next next part

redPanther:
now: separate init code from write. Not yet perfect, but it's a start :-)
"setConfig" is renamed to "init"

no


penfold42:
You've been a busy boy again - great work, thanks !

redPanther:
Thx for the flowers.
I tried to separate it in a thread. Just threading is fast done. Problem is that qt communication classes (network, serial) are not so easy to handle in a thread. The signal/slot system needs care, because running in a thread and use eventLoop from the main thread isn't possible.
If you find working examples for network and serial stuff in a thread, I'd would be thankfull

Issue #59, opened 2016-06-27
Colors: Fix minLuminance, add enable/disable option (also during runtime)


brindosch:
Option: Set the min luminance just during grabbers are active, so it will be diabled if you use a "idle effect/color" or just black with disabled grabbers.
For this you may "move" this to another(new?) component which could be also called (started and stopped) during runtime.

Fix: my ambient light get strange colors at black areas. Looks weird. Another solution is appreciated.
Hint: The old deprecated value "blacklvl" works like a charm, but might be more a hack. As long it works?

redPanther:
color transformations can't be started/stopped. You can think like that that:

- grabbers are like a cd-player, there you can start/stop/pause
- effect engine are like a tuner (you can select your "programm")
- color transforms are more like the equalizer. You can bypass it, but not start/stop
- color smoothing is more like a dsp that can make nice echo/reverb effects. (bypass only)
- all things go into the priorityMuxer. This is somehow like the channel switch on the amp (tuner,phono,cd,aux), but with detection if a channel plays music. if not auto switch to next input source
- The leddevice is like an amp
- and the leds itself are like the speakers.

in conclusion you can start/stop only components that produces images or led colors (setimage/setcolors)

Is your wish that luminance settings only applied to grabbers? ATM we can't decide from where the colors comming from and in a way we will never do. But we can change the place where this specific settings are applied. One possible solution: apply this directly after the grabber before it goes to muxer - downside: this works as long as protoserver is only used for grabber data - or better say, we can't decide if it comes from proto or from grabber and which type is proto sender.
Currently we have only grabbers as proto sender ... 

ATM we don't have "bypass" for our transformations.  In our hifi-set picture means that:
If you don't want the EQ then disconnect all cables and throw it away. If you need it, buy a new one and connect it. This is a bit to dramatic - a bypass switch in  the EQ would perform better :-)
This means e.g. for color smoothing. This component should always be in our chain. When "disabled" (=bypassed) then this component write all input data directly to the output.

penfold42:
The more I think about it, I'm thinking we need 2 sets of transforms.

One for the LEDs and one for the grabbers.

Each have specific reasons for needing some transform and with multiple sources and multiple receivers a single transform will get confusing 

On a related topic - colororder.
This probably needs to move from the core and be pushed closer to the led devices

redPanther:
You mean input dependend transform settings? We have only one output so no need for specific output settings.
Currently transforms are last step before led output.

[...]

PR #267, opened 2016-10-09
implement optional color correction for V4L only

redPanther:
Now it is possible to have color adjustment and color transform only for v4l input.
This is needed, because usb grabber often need heavy color adjustment, while effects, solid color or screen capturing often needs no correction.
Now you have the choice!

_additionals:_
 hyperion update function now knows from which component the current color comes. Perhaps this info might be handy in time ...
color temperatue is removed ... I don't like duplicated code!

webui: adapt config write to latest qjson PR. This function will always commit full config, regardless if your input is full config or a part of it.

2 new values in color section:
		"channelAdjustment_v4l_only" : true,
		"transform_v4l_only" : true,

look default configs for details.
I set default to "true", so by default color correction is only for v4l. I think this is the most usecase ...

and: **color temperatue is removed**. The same thing can be achieved with color adjustment

#59


brindosch:
amazing again :)
note: hyperion-remote cleanup is missing (temperature)
Q: Now if you add a second transform object into the array with splitted leds (0-50 and 51-80) does it still work to assign it properly to v4l2 only?

> I think this is the most usecase ...

Hm, hard to say. As long Openelec is the download count leader i bet against it ;)

redPanther:
if you assign v4l_only any color object will be v4l only. Currently you only have the choice to assign to all input sources or only to v4l.
hyperion remote is now cleaned. There was also the old correction ... I removed it too, because it was the same as temperature.

redPanther:
... and changed defaults to v4l_only false - this is the same behavior as we have now.

Paulchen-Panther:
Auch von mir ein großes Danke für diesen PR. Respekt.

PR #268, opened 2016-10-10
webui: add read/write for all editors

redPanther:
as the title said .. add read/write for all editors in webui.
Now you we can write all values (except log and webui - I decided that are 'nerd' settings and they can only be set in config manually)

when trying out, please backup your config before!!! (just some precautions ... )

no


redPanther:
... a side note layout and beautiful eye candy wasn't the goal of this PR - just the basics of loading and writing the config

b1rdhous3:
Nice! But how about a "nerd"-Mode on webui?

redPanther:
of course, but  the json editor needs some extension.
I want something like "settings level". Then we can add a settings level in json schema and set the level on editor ....

brindosch:
and a title catcher(compare) for translations? the only idea i have how we could integrate them?

PR #283, opened 2016-10-31
update configurator

brindosch:
Now with "start effecttest" "stop effecttest" and " toggle continuous test" buttons and some cleanup.


redPanther:
great stuff thanks!

PR #285, opened 2016-11-09
Fixed QHostAddress address binding for AtmoOrb

RickDB:

QHostAddress::Any no longer works with used Qt version, this resolves it by changing it so that it binds to Any IPv4 address.


brindosch:
Thank you!

PR #286, opened 2016-11-14
Updated compile instructions for Mac OS

rudelm:
I've updated the compile instructions for Mac OS. I've tested them on Sierra and El Capitan.

Note: For further discussions use our forum: forum.hyperion-project.org

PR #294, opened 2016-11-20
platform cleanup and cmake osx fix

redPanther:
- fix cmake platform logic for osx. now 
```cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_BUILD_TYPE=Release ..
make -j4
make install/strip```
 should be enough - optional add CMAKE_PREFIX_PATH and CMAKE_OSX_SYSROOT if neccessary

- remove platform for rpi (for pi 1)
- platform rpi-pwm (pi 2/3) is renamed to rpi
- remove last pieces of WS2812B PWM driver
- add detection for odroid-c2 (experimental, because can't test)
- enable tests for travis osx build
- install hyperion binaries to as symlinks additionally to bin folder. then they are automatically in system path
- create effects folder for custom effects (with readme with some infos)

no

 #286


penfold42:
I've found on my osmc pi3 box that I'm running out of ram on higher multi CPU builds.

Is it just me ?

Should we look to cap nprocs on the Pi ?

I have no swap space defined to that clearly isn't helping my cause, but if the build has to swap it'll probably run even slower!

redPanther:
the amount of jobs is more or less a suggestion to speed things up. It is up to everybody to use it's own value here. E.g. when you use distcc you can spread your compile job on multiple pi's :-) you will set higher values.
But we should mention the "out of memory" possibility in the compile howto.

Issue #284, opened 2016-11-04
Adalight fastled arduino file mod


iLLiac4:
Do someone knows how to implement this part of the code into this ino file.
The file that needs to be modified:
[https://github.com/hyperion-project/hyperion.ng/blob/master/assets/firmware/arduino/adalight/adalight.ino]

With the logic from this file [https://github.com/iLLiac4/Adalight_Hyperion/blob/master/Adalight_WS2801_Mini_Pro_32u4.ino] that uses SPI.h logic which turns off the lights if no data is received for some time.

`      // If no data received for an extended time, turn off all LEDs.
      if((t - lastByteTime) > serialTimeout) {
        for(c=0; c<32767; c++) {
          for(SPDR=0; !(SPSR & _BV(SPIF)); );
        }
        delay(1); // One millisecond pause = latch
        lastByteTime = t; // Reset counter
      }`

iLLiac4:
Ok i did it so you can use it and test it.
https://github.com/iLLiac4/Adalight_Hyperion/commits/master/experimental.ino

redPanther:
I did some changes to our version inspired by your file:
https://github.com/redPanther/hyperion.ng/blob/bugfixes/assets/firmware/arduino/adalight/adalight.ino

I'm preparing a pr, if you want to test the sketch ...

iLLiac4:
I have tested it and it seems to work.

 

There is also one ino file in my repo that can use adalight protocol and ordinary rgb led stripe. I have named it poors man ambilight because it will cost no more than 10usd. It is meant for arduino. https://github.com/iLLiac4/Adalight_Hyperion/blob/master/Adalight_RGB.ino

The wirinig can be the same as http://www.forum-raspberrypi.de/Thread-hyperion-tutorial-ambilight-mit-standard-led-strips but this one is not for RPI but for Arduino.

 

http://www.forum-raspberrypi.de/attachment.php?thumbnail=5870

iLLiac4:
Also yesterday i have found premade wifi controller http://www.banggood.com/ARILUX-AL-LC01-Super-Mini-LED-WIFI-Smart-RGB-Controller-For-RGB-LED-Strip-Light-DC-9-12V-p-1058603.html?p=VD18093452842013112L  (it has to be from 5v-28v) that is based on esp2866. I am planning to mod it to accept ordinary stripes rgb and also ws2801 or. Apa102.

[...]

PR #299, opened 2016-11-27
update adalight sketch

redPanther:
- add analog output to adalight sketch 
- add basic color corrections to adalight sketch
- add possiblility to switch of leds when no new data is comming in
- some refactoring of that sketch

#284 


redPanther:
@penfold42 this is needed when smoothing is off. smoothing output is mostly fast (in ms interval), the serial refresh time is more a "keep alive" and has an interval in seconds

penfold42:
Cool - we should probably consider doing this for all led device types ?

redPanther:
Good idea. The code for refresh is not new in rs232. I made it available in webui.
I will move it to leddevice baseclas by time.

redPanther:
What l havn't mentioned, the new adalight sketch make it possible to drive addressable and analog stripes at the same time.

redPanther:
Hi, last days I puzzeled with my new sketch. The analog was flickering. I've thought my sketch was to slow or pwm interferes with fastled - yesterday I found the cause. Thx to my scope, I found out that some pins have 'faster' pwm then other and they are out of sync!
Here some info on that topic http://playground.arduino.cc/Code/PwmFrequency

I will update my sketch according to this new knoledge

PR #305, opened 2016-12-01
refresh time for all devices + schema fixes

redPanther:
as the title said ...

new general led device opt: rewriteTime


brindosch:
Ouh thats weird for me now hehe.
https://github.com/brindosch/hyperion.ngBeta/commit/c5cb3ab56a0e5f4d9efb876b1d8eacbffdb8bb9f

redPanther:
hmmm so we have to merge .... but now leddevice config is more complete. There were some device options missing due to misspelling

brindosch:
Should this setting be hidden?

penfold42:
I don't think so.

What would be cool tho is "advanced" and "expert" level options that the web up could selectively expose

b1rdhous3:
We could add levels to the schema so we can hide and show some options and activate them on webui with a button ("advanced" and "expert")

brindosch:
yeah of course, should be no problem
We need to discuss the settings storing, easiest would be inside our .json. But i don't like the idea because:
 - a level option/language setting and so on are not bound to a single config.json.

These should be stored and (json rpc) updated immediately. Could be also used in applications to setting up a language

brindosch:
It's done. How many levels do we need (2? 3?) And if so, which editor input should be just available at a higher level. I need a list :)

redPanther:
for my taste: 2 levels like VLC - "simple" / "all"

penfold42:
Agree - just 1 set of hidden advanced options.

PR #309, opened 2016-12-04
Transit translation system

brindosch:
 - As we are using html tags and js based translation i changed the lib to this one https://github.com/wikimedia/jquery.i18n
 - Created an en.json with ALL english strings, this is necessary for a proper working fallback
 - alle schemas no longer use hardcoded titles and appends, instead everything runs through the translation step
 - "edt_dev" = led device related, "edt_conf" = config related, "edt_eff" = effect schemas related, "edt_append" = all unit stuff, "edt_msg" = all internal validaten messages of editor + buttons.
 - all editor elements translated to DE
 - Currently the lib use your bowser language for translation
 - you could no longer change your language during runtime as the js based stuff is done when it's done, a browser reload would be the minimum task to enable it again and store the language in a cookie or in html5 local storage.
 - effconfigurator optimizations from @Paulchen-Panther . Buttons are now disabled if you shouldn't click them and all schemas are loading properly now
 - inside i18n/qqq.json you find a first part of a translation string docu. i consider to use https://translatewiki.net/ for translation or a similar service
 - logging page reworked with log level setting and auto scrolling enable/disable
 -  enjoy


redPanther:
impressive changes!!!

brindosch:
Thank you, i have already an idea how to implement language selection very smooth.
the default setting is "auto" when auto, it uses your browser locale. If this won't work for you or you want another language you could set it to "en", "de".... A Dialog "set your language" isn't necesssary anymore.

redPanther:
as storage I would prefere webstorage http://www.w3schools.com/HTML/html5_webstorage.asp
(as long as we have only a few ui settings - currently only language and settings level (easy/expert) )

brindosch:
I am not the greatest fan of this, cause if you switch device you need to re configure it again (if not "auto"). So i thought rather storing this in webConfig part to ship around this. You are always just one hyperion restart beside.
Simple comfort problem anyway.

b1rdhous3:
So we need new "general settings" schemes?

brindosch:
I don't think so. Hide danger settings in "expert" and you are fine to show webConfig part without the chance of messing around.

b1rdhous3:
And add language and "view mode" to the webConfig settings?

brindosch:
Did so.
unbenannt

b1rdhous3:
Very nice!

brindosch:
After i tested a little bit and thought about pro/con. I will just use html storage. The impact of loading time is high to wait until hyperion response and a hybrid solution doesn't work across diff browsers.

redPanther:
browser storage rulz :-) then I can have different settings on desktop and tablet :-)

Issue #204, opened 2016-08-24
Enable/Disable components during runtime - Issues


brindosch:
- [x] Blackborder
- [x] Smoothing
- [x] Kodicheck
- [x] v4l2
- [x] Grabber
- [x] boblight
- [x] udplistener
- [x] forwarder

More not tested yet

Feel free to use this PR for easy testing:
https://github.com/hyperion-project/hyperion.ng/pull/203

redPanther:
and the last 3 services? any testers? ( @tpmodding  ) for boblight and udp listerns you can use nmap or other network port spy tool to see if hyperion opens the ports or not ...

tpmodding:
i can test boblight and forwarder

redPanther:
.. that would be great :-)

@penfold, can you test if udplistener vsn be enabled/disabled through hyperion-remote and/or webui?

penfold42:
It seems really buggy via web interface

most times i get "Error while parsing json: illegal value at Line: 0, Column: 2" (or some other line/column reference).
The one time it did save without error, it did rewrite the config file, but the change didnt take effect until i restarted it.

Also noticed this:
[HYPERIOND JSONCLIENTCONNECTION] <ERROR> Someone is sending very big messages over several frames... it's not supported yet

penfold42:
./hyperion-remote  --disable UDPLISTENER
./hyperion-remote  --enable UDPLISTENER

both seem to work fine

redPanther:
@tpmodding any news from boblight and forwarder? I would like to close this issue

tpmodding:
total vergessen, ich tests heute, ich geb dann feedback

brindosch:
> ich tests heute, ich geb dann feedback
:)

tpmodding:
jap bin dabei, hab probleme mit boblight... forwarder funzt zwar aber boblight macht mucken...bin aber nicht sicher obs an hyperion oder am boblight liegt der daten verschickt bzw nicht verschickt,
auf android geräten mit kodi bekomme ich kein boblight installiert, der will abhängigkeiten aber spuckt dabei fehler aus (kodi auf fire tv stick). muss jetzt mal nen neues libre installieren, bin noch auf einer alpha

tpmodding:
@redPanther i updated everything today again and now my leds dont light up... if iy use hyperion it works, hyperion.ng workds nothing anymore?

brindosch:
[...]

PR #315, opened 2016-12-05
fix for LPD8806 issue

penfold42:
fix for LPD8806 issue - wrong number of bytes sent on setup
make writebytes call consistent with other devices

no

#204 

### DIFF
--- a/libsrc/leddevice/LedDeviceLpd8806.cpp
+++ b/libsrc/leddevice/LedDeviceLpd8806.cpp
@@ -21,7 +21,7 @@ bool LedDeviceLpd8806::init(const QJsonObject &deviceConfig)
 	_ledBuffer.resize(messageLength, 0x00);
 
 	// Perform an initial reset to start accepting data on the first led
-	return writeBytes(clearSize, _ledBuffer.data()) >= 0;
+	return writeBytes(_ledBuffer.size(), _ledBuffer.data());
 }
 
 int LedDeviceLpd8806::write(const std::vector<ColorRgb> &ledValues)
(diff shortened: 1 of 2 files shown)


libsrc/leddevice/LedDeviceLpd8806.cpp
```diff
@@ -21,7 +21,7 @@ bool LedDeviceLpd8806::init(const QJsonObject &deviceConfig)
 	_ledBuffer.resize(messageLength, 0x00);
 
 	// Perform an initial reset to start accepting data on the first led
-	return writeBytes(clearSize, _ledBuffer.data()) >= 0;
+	return writeBytes(_ledBuffer.size(), _ledBuffer.data());
```
redPanther: I'm not sure if we should replace clearsize, because original driver does the same:
https://github.com/hyperion-project/hyperion/blob/master/libsrc/leddevice/LedDeviceLpd8806.cpp#L30

penfold42:
I agree - was looking at this on my phone and had my browser tabs mixed up !

redPanther:
.. I know that problem well :-) BTW I think you know, but to be sure, here is the explanation for the driver https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/leddevice/LedDeviceLpd8806.h

redPanther:
I've merged and changed back the clearsize

penfold42:
Yep - read that and then misread it in the context of my browser tab mix up *laughs*

It should be a harmless "error"
Let's see what tpmodding comes back with and then I'll revert just that error if the other changes work

penfold42:
Just saw your last message - thanks

PR #317, opened 2016-12-07
Added new Candle effect - like those fake flickering led candles

penfold42:


brindosch:
Oh, sounds nice.
Do you want to add a effect schema?

penfold42:
I do, but right now it can parse the effect args looking for a list, a number or a string.

Does the schema support multiple data types ?

brindosch:
You mean for example a integer or string for the same option? Not yet. But would be hard to display this in an ui. As you never know when the field should be type=number, type=text, or a select. A meaningfull input type helps a lot. As numbers have spinners for mouse cursors and on touch the number keyboard opens instead text. Selects open a modal to select from instead typing text in a textfield.

brindosch:
```
+# candles can be :
 +# a single led number, a list of candle numbers
 +# "all" to flicker all the leds randomly
 +# "all-together" to flicker all the leds in unison
```
So the solution for this would be something like this:
select : all, all-together, single
text : single led(s)

if (single) use textfield.val()
Connecting code and ui 1:1 could have weird results. I think there is something like dependency available to hide options. There are other effects that should make use of it too.

It is possible that we fail during schema check, if there is something missing. Not sure. Will test

penfold42:
Agree on the solution - if the text field can be empty from a schematic perspective we're all good.

I can parse the text field for commas or spaces in python

effects/candle.py
```diff
@@ -0,0 +1,66 @@
...
+color = hyperion.args.get('color', (255,200,0))
+sleepTime = float(hyperion.args.get('sleepTime', 0.14))
+brightness = float(hyperion.args.get('brightness', 0.5))
+candles = hyperion.args.get('candles', "all")
+
+candlelist = ()
+if ((candles == "all") or (candles == "all-together")):
+	candlelist = range(hyperion.ledCount)
+elif (type(candles) is int):
+	if (candles < hyperion.ledCount):
```
brindosch: candles <= hyperion.ledCount may be better

brindosch:
There is no form of hide/show fields depending on another fields value. Funny that such a mandatory function is not available. But crazy support to watch fields and copy content of them.

effects/candle.py
penfold42: Noooo

10 leds are numbered from 0 to 9

penfold42:
I'm nearly done on the schema - most is eorking.

where is the l10n catalog ?

penfold42:
Merge me please ?

brindosch:
Would be hard with this distance :)
Ah! wonder why police stopt working

brindosch:
So after some testing...
Very nice :+1:
Would appreciate more of this. (As our effect list regarding 'mood' is not that long and some of these effects are just weird :D)

penfold42:
Thanks for the feedback, Glad you like it!

It got me thinking we needs some online help... I'll open a separate issue

PR #329, opened 2016-12-10
Update CompileHowto.md to say that Ubuntu instructions apply to Win10 Linux Subsystem

XfableX:
Me and penfold42 compiled it on Windows 10 Bash window and it compiled and ran, so I added it to the compile howto

### DIFF
--- a/CompileHowto.md
+++ b/CompileHowto.md
@@ -1,6 +1,6 @@
 # Install the required tools and dependencies
 
-## Debian/Ubuntu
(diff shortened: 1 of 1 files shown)


IIIdefconIII:
please document the correct cmake command for windows machine 

Keep in mind to disable all linux specific led and grabber hardware via cmake. Because we use QT as framework in hyperion, serialport leds and network driven devices could work.

Issue #331, opened 2016-12-10
Reserve a Prio-Range for WebUI


b1rdhous3:
I've set my v4l2-grabber to prio 1 and wasn't able to set colors or effects from webui because they all use prio 1.

So just a thought: Shall we reserve a range of prios for our internal stuff?

redPanther:
We schould limit settings to higher 1 and lower max int -2. But only in webui. If somebody does this willingly with a editor in config file ....
 we can't prevent that people doing stupid things.
Perhaps setting min/max in grabber schema do the job. Then also manual config edit is catched

tpmodding:
the idea with the webui limit is good! we can not decide which priority have to use others :)

brindosch:
Don't mess around with priority!
That's it https://github.com/brindosch/hyperion.ngBeta/commit/306400536920c2b7af720d296e792ec936838c15

Issue #330, opened 2016-12-10
Windows 10 linux subsystem limits


XfableX:
I have recently compiled Hyperion off of windows 10 using the Linux subsystem.
So far I know it works over wireless LED devices, the web interface works and the app works but have yet to find what does not. So far I have found that the Subsystem cannot see USB ports which will be a limitation.
Can more people try and discuss their findings with the subsystem?

tpmodding:
awesome. did not know that its possible! windows system should work with com-port connections and an arduino?

redPanther:
We use qt, it should be possible with some modifications regarding #include to compile with visual studio. But the big question, who wants to make the support job.

XfableX:
Unfortunately I haven't really used visual studio at all but I can give it a try.

redPanther:
I will close this issue ... we know now the it basicly works. Windows is currently not our development focus. More discussions to that topic can be  done at https://hyperion-project.org/

Issue #292, opened 2016-11-19
Config write lead to closed websocket


brindosch:
A config write request:

```
Nov 19 16:40:00 osmc hyperiond[706]: [HYPERIOND JSONCLIENTCONNECTION] <DEBUG> <JsonClientConnection.cpp:200:doWebSocketHandshake()> Websocket handshake
Nov 19 16:45:20 osmc hyperiond[706]: [HYPERIOND JSONCLIENTCONNECTION] <ERROR> Someone is sending very big messages over several frames... it's not supported yet
Nov 19 16:45:20 osmc hyperiond[706]: Someone is sending very big messages over several frames... it's not supported yet
Nov 19 16:45:20 osmc hyperiond[706]: [HYPERIOND JSONSERVER] <DEBUG> <JsonServer.cpp:64:closedConnection()> Connection closed
```

I never seen this before on save. I just wanted to implement the led layout writing.

```
	$("#leds_custom_save").off().on("click", function() {
		function createLedConfig(){
			var string = '{"leds" :';
			string += $("#ledconfig").val();
			string += "}";
			return string
		}
		if (validateText())
			requestWriteConfig(JSON.parse(createLedConfig()));
	});

```
There should be no difference as we write the entire config _always_ 
So what happend now?

brindosch:
Noticed we need a config buffer as you could _really_ just save once per session :)

redPanther:
can you test, should be working now with around 600 leds (I got 630 to work)

redPanther:
> Noticed we need a config buffer as you could really just save once per session :)
now you can save as many times you want per session

PR #333, opened 2016-12-13
detect config is readonly, adalight and other stuff

redPanther:
- update adalight sketch (this won't be the last)
- update lightberry sketches to latest upstream
- add tpm2 sketch. mainly for dev's so we can test tpm2
- tune ada driver, set delayAfterConnect default to 1.5s because some arduino (e.g. mega r3) needs this
- update compilehowto (windows disclaimer)
- some refactoring in main cmakelists + preparation for windows compile
- set priority min/max for grabber/network services - prevent colliding prios between webui/background stuff and grabbers/net services
- add check if config is writable. TODO do something usefull in webui
- fix webui can't write led config

no

#330 #331 #292


assets/webconfig/js/content_leds.js
```diff
@@ -472,11 +472,8 @@ $(document).ready(function() {
 	
 	// ------------------------------------------------------------------
 	$("#leds_custom_save").off().on("click", function() {
-		function createLedConfig(){
-			var string = '{"leds" :'+$("#ledconfig").val()+'}';
-		}
 		if (validateText())
-			requestWriteConfig(JSON.parse(createLedConfig()));
+			requestWriteConfig(JSON.parse('{"leds" :'+$("#ledconfig").val()+'}'));
```
brindosch: uhm, of course. That was my intention. For a unknown reason i stopped in between :)

brindosch:
Nice!
Will do the ui stuff.

redPanther:
> Will do the ui stuff.
I was  hoping you will do. BTW I tested webui on my x86 and my pi1 both can write the config. (tried my standard config with 116 leds)

brindosch:
PLATFORM: rpi

ENABLE_FB = OFF
ENABLE_SPIDEV   = OFF
ENABLE_V4L2 = OFF

Hm, no :)

src/hyperiond/hyperiond.cpp
```diff
@@ -72,9 +72,9 @@ HyperionDaemon::HyperionDaemon(QString configFile, QObject *parent)
 		WarningIf(_qconfig.contains("logger"), Logger::getInstance("LOGGER"), "Logger settings overriden by command line argument");
```
brindosch: overridden :)

brindosch:
The notifications seems not to work for me. I edited the write access and restartet always. But it still hangs in writeable : true

-r--r--r-- 1 osmc osmc 20553 Dec 14 00:23 hyperion.config.json

b1rdhous3:
Ich hab den Adalight-Treiber nun mal mit Lightberry HD und dem USB Connector getestet. 
Läuft alles! :)

src/hyperiond/hyperiond.cpp
redPanther: we should make a text typo hunt ;-) I'm sure we have more. Perhaps a english native can help ;-)

redPanther:
@brindosh you are running as root, than all files are writeable ragardless of the permissions. I use hyperion as user, then this check works.

brindosch:
Ah, wonder! I will try as non root again :)
I am still slightly confused, as i had the hope this was the reason why i can't save configs.
Will test futher.

redPanther:
Saving: I figured out that I can save only once per session. That means Ihave to reload the ui. Now I can reproduce and figure out waht's the reason.

Btw I successfully created my special led layout with 96 ws2812 leds + 4 ikea dioder stripes with the webui. What I missed is an option to append generated config to current, instead of replaceing it.

brindosch:
> What I missed is an option to append generated config to current, instead of replaceing it.

append a led layout to the current? I can't imagine something different.
I thought about a backup/restore function. but we need a version key in config that is maintained from Hyperion itself to keep old configs away

PR #336, opened 2016-12-17
Re-introduce hsv transform into config schema.

penfold42:
iOS app needs this (at the moment) and i needs to be able to set the inital config from the config file.

```
                "transform" :
                [
                        {
                                "id"   : "default",
                                "leds" : "*",
                                "hsv" :
                                {
                                        "saturationGain"   : 1.0000,
                                        "valueGain"    : 0.1000
                                },
```

I need it!! thats an issue *grin*

### DIFF
--- a/libsrc/hyperion/hyperion.schema.json
+++ b/libsrc/hyperion/hyperion.schema.json
@@ -233,6 +233,23 @@
 								"type" : "string",
 								"required" : true
 							},
+							"hsv" : {
+								"type" : "object",
+								"required" : false,
+								"properties" : {
+									"saturationGain" : {
+										"type" : "number",
+										"required" : false,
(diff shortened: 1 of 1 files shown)


redPanther:
thx, I also thought of that. Now everybody can choose what's the best ....

brindosch:
I am not that much with you guys
Sometimes you don't want to choose, it should just work.

And mixing/display both has bad side effects and you are 'lost in settings'

A proper ui is mandatory to keep it

penfold42:
Why ? We can deprecate it for now exposed only via json, Hyperion-remote and the config file without cluttering the web UI.
Or make it an "advanced" feature in the web UI.

When the iOS app is updated and the HSL bugs are fixed (issue coming soon) we can nuke it.

Issue #10, opened 2016-06-17
adjustable image to led mapping


redPanther:
Hyperion calculates  average color of rectangular screen area to a led. Every led has his area.
Currently calculating average color of whole screen and set this single color to to all leds is not possible. (ok there is a trick, but this waste to much cpu time)
This is one thing ... the other is, not everybody likes average color. There are other algorithms like weighted average, mean vaule, [dominant color](https://hyperion-project.org/threads/output-color-determination-algorithm.117) 

Hyperion needs some transform mechanisms for image to led mapping. We already have "color transform", but this make the transformation after the image processing.

redPanther:
this issue is more or less related to this
https://github.com/hyperion-project/hyperion/issues/72

brindosch:
You are bored? :)

redPanther:
just looked around for v4l things and found that by accident ...

found that: https://github.com/b2un0/kodi-script-hyperion/issues/3
It's not a big problem to add multiple color adjust/transform with only one config. (just a bit monkey work)
We can have a default "color" section and "color" sections per prio-channel. When e.g. v4l is on 901 then make a "color" section with "prio channel assignment" "901" and if this channel is active, then this config is selected and if any other channel is active, take the default ...

brindosch:
possible :)
sounds nice - optional :)

tpmodding:
so we can have for different sources different color adjustments?

brindosch:
the unicolor mode maps from the entire picture one color to all leds, multicolor is the old behaviour.
a per priority mapping of color adjustments (which are not part of the headline for this issue) are not available (yet)

redPanther:
Not exactly, we can limit transform and/or adjustment to v4l source

PR #341, opened 2016-12-18
adjustable image2led mode for grabbers

redPanther:
you can switch between multicolor (all leds could have different colors) und unicolor mode.
unicolor means all leds have sa same color. The color is calculated from the average of the whole(!) image.
This method is not the best, because it gets sometimes to dark, but it's a start. This methode is very cool when you watch pictures.
In general we have now a methode to implement other image to led mapping algorithms.

usage:
use it via hyperion-remote or json interface with you own application:
unicolor mode
```
bin/hyperion-remote -M unicolor_mean --print
hyperion-remote:
        Version   : 2.0.0 (unicolor (redpanther-90f27a1/cbec2f4-1482083992))
        build time: Dec 18 2016 19:02:32
Connected to: "localhost:19444" 
Command: {"command": "processing","mappingType": "unicolor_mean"}
Reply: {"command": "processing","success": true,"tan": 0}
```

multicolor mode:
```
bin/hyperion-remote -M multicolor_mean --print
hyperion-remote:
        Version   : 2.0.0 (unicolor (redpanther-90f27a1/cbec2f4-1482083992))
        build time: Dec 18 2016 19:02:32
Connected to: "localhost:19444" 
Command: {"command": "processing","mappingType": "multicolor_mean"}
Reply: {"command": "processing","success": true,"tan": 0}
```

you can set default mode in hyperion.config:
```
	"color" :
	{
		"imageToLedMappingType" : "multicolor_mean",
```
or "unicolor_mean" ...

 #10

### DIFF
--- a/config/hyperion.config.json.commented
+++ b/config/hyperion.config.json.commented
@@ -30,6 +30,8 @@
 	/// Color manipulation configuration used to tune the output colors to specific surroundings. 
 	/// The configuration contains a list of color-transforms. Each transform contains the 
 	/// following fields:
+	///  * 'imageToLedMappingType'      : multicolor_mean - every led has it's own calculatedmean color
+	///                                   unicolor_mean   - every led has same color, color is the mean of whole image
 	///  * 'channelAdjustment_enable'   : true/false enables/disables this channelAdjustment section
 	///  * 'channelAdjustment_v4l_only' : if enabled and set to true, then channelAdjustment is only for v4l devices
 	///  * 'channelAdjustment'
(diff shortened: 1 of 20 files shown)


b1rdhous3:
So we should add a simple button in the webui remote for this?! =)

redPanther:
somewhere in the remote control would be fine. the settings for the config is already there

brindosch:
I am to slow or you to fast! I will do... :)

brindosch:
Regarding feedack, the same "ui issue" i have with the auto selection button. Could we output 'somewhere' the state to set a proper button styling

Do you want to add the auto mode (source selection) like the "off" mode to the list to get proper state feeback?

Thank you :)

redPanther:
I realized already thst i forgott to expose the state to json api. A function alredy exists, but have to be added to json serverinfo

teeedubb:
Hi redPanther, thanks for this addition, this is one feature which I missed from my old setup. 
One question, is it possible to make this feature work with the kodi addon (which uses the UDPLISTNER input)? Toggling unicolor_mean when it is active has no effect. Work wonderfully with the v4l grabber.

Thanks again for this PR.

PR #343, opened 2016-12-19
impl options explanation, uiLock

brindosch:
 - Adds optional explanations for basic conf. Auto generated (could be never out of sync)
 - uiLock if config is not writeable
 - editor: fixes empty h3 root editor header
 - page introduction is now hidden on small device screens
 - led layout is table based instead formgroups, fixed initial led preview
 - editor: labels and their inputs are now connected (click on label highlights the input field or change state of checkbox)
 - editor: if a default value is available, it will be added as a placeholder
 -  remote: Added led mapping control and state reflection for auto select button
 - don't requestLoggingStop if logging is disabled
 - Avoid effectnames that begins with a dot


brindosch:
regarding prio: i suggest a range from 2 to 99 for remotes and everything above is capture stuff.
watchdog reset now every 8 seconds, with bad wlan connection much better now! Well, side effect is a delay for reload if the interval triggers during a real lost connection

libsrc/jsonserver/JsonClientConnection.cpp
```diff
@@ -486,9 +486,9 @@ void JsonClientConnection::handleCreateEffectCommand(const QJsonObject& message,
 				
 				if (effectArray.size() > 0)
 				{
-					if (message["name"].toString().trimmed().isEmpty())
+					if (message["name"].toString().trimmed().isEmpty() || message["name"].toString().trimmed().cbegin() == ".")
```
redPanther: || message["name"].toString().trimmed().startsWith(".")

brindosch: Ah! Next time i try the js way instead something different ;)
Is there somewhere a good overview with good examples for beginners?

For js i found the mozilla developer network. Very good to understand the basics and how they behave, list of possibe functions etc. The QT5 doc feels a step too high :)

redPanther:
merge when you ready :-)

brindosch:
The following changes and additions have been made
 - editor: labels and their inputs are now connected (click on label highlights the input field or change state of checkbox)
 - editor: if a default value is available, it will be added as a placeholder
 -  remote: Added led mapping control and state reflection for auto select button
 - don't requestLoggingStop if logging is disabled
 - Avoid effectnames that begins with a dot

Issue #328, opened 2016-12-10
Move led device name to be a global instance name?


penfold42:
This should really be an attribute of the Hyperiond instance and not the led device.

Thoughts ?

redPanther:
I thought this from my first hyperion experience ... I prefer we make a new section "general" and move the name (rename it to hyperionID?) into it. We could also move the leg level into that section ... Later on we can add something like "update notification" and "update notification mail" to send a mail, when a new version is out. ....

brindosch:
> leg level into that section

I think we should keep the loglevel where it belongs to :)
Email notifications are a good idea.

redPanther:
close?

PR #345, opened 2016-12-21
move 'name' to new config section general.

redPanther:
Move name to new section general.
Add version number to config to describe the schema version.
Prepare structure for config migration feature

New section general and and a version number for config

#328 


brindosch:
Nice one.
Do you want to compare schema 1 against schema 2 for migration?
In case - you skip a version you can't migrate and you could never go back?
Isn't it possible to delete keys and add keys from config file of they are in schema/not in schema?

To hide the version we could add new access level that is always removed. "system" or such

redPanther:
Good idea. Can you pr the new settings level. I will integrate it.

redPanther:
Migration:
- Load and validate current cfg
- get current version of config
- if this is what hyperiond needs, exit
- otherwise start function to migrate from current to next higher version
- do this again, until we reach needed version, then exit

Everytime we increase the version number we have to write a function to migrate from prior version to the new.
This is the savest way.

brindosch:
https://github.com/hyperion-project/hyperion.ng/commit/be909813880de33a16eddb2448234d69c1b29510

PR #346, opened 2016-12-22
Trails effect

jonudewux:
Tested on 32/8 led matrix.


redPanther:
Thx a lot. Could you supply a schema file too? It is located in effects/schemas.
This file is needed for creating custom effects via webui

Issue #276, opened 2016-10-17
Color calibration discuss


brindosch:
As we slowly reaching a interesting step:
There are 3 types of use cases - we should catch them all
- Use just framegrabber or other "digital" sources (digital calibration)
- Use framegrabber and v4l2 (digital calibration that is applied to _all_ and a second calibration that is used for v4l2)
- Use just v4l2 (calibration for v4l2 only?!)

Additional question:
Effects and Colors - Should they skip the color modifications entirely? Does it harm someone with strange colors for effects/colors if they are not processed? (If i see people having 50,255,50 to get a white color?!)

Additional task:
merge 2 arrays in 1. Adjust transform and channeladjustments separably is crazy to use.
As code == gui i suggest a new sorting that follows the configuration flow and not the code flow.
See HyperCon: https://hyperion-project.org/wiki/HyperConProcessTab-Colors-EN

And on top:
Splitting smoothing between colors/effects and all other stuff (own setting - i think my smoothing is too high for effects, but good for pictures)

HSL vs HSV
Could you use HSL to lower the overall brightness without getting very weird colors (compared with HSV)?

redPanther:
for effects/color "transform" is not usefull, but "adjustment" is needed. Some leddevices (fadecandy/adalight), can do same operation in hardware, so for that devices the settings not necessary - other devices e.g. all stripes directly driven on rpi gpio need this settings.

I think first task is to merge color settings in one settings block. Then we will see whats next.

penfold42:
Is like to see separated settings
1) grabber settings to correct shitty grabbers
2) led device settings to correct wacky leds

redPanther:
when I think about, should we stick "channeladjustments" to leddevice and "transform" to grabber.

When doing calibration set "channeladjustments" with solid colors/effects.
Then start grabbing and optimze result with "transformation".

Did I miss something? We can do this already. 
Just set:
```
"channelAdjustment_v4l_only" : false,
"transform_v4l_only" : true,
```
what we don't have is a "transform" for screen grabber (x11,dispmanx,..) 

perhaps we should:
- remove `"channelAdjustment_v4l_only" : false,` this should be active for all
- replace transform_v4l_only with (e.g.) "transform_target" with following possible values (enums) `all`, `v4l`, `screen capture`
- possibility to have "transform" block 2 times, so you can have different settings for v4l and screen capture

brindosch:
So current state:
[...]

PR #351, opened 2016-12-26
Color Adjustment 2.0

Funatiq:
See [forum post](https://hyperion-project.org/threads/color-adjustment-2-0.552/).

See config files below.

To do:
Hyperion remote needs to be updated for the new colors: black, white, cyan, magenta, yellow


brindosch:
Looks promising, some thoughts. As you see the color area is still very untouched in comparison to current stable and needs a 'refactoring'. The way we have chosen how to connect ui and and config files is neat but has also some limits.

 - I think it's suitable to use a colorpicker to modify each color?
 - you could rename pureR/G/B accordingly. We break regurlarly the configs anyway :)
 - The "black" could be removed from config? (Any downside possible?)
 - Did you succesfully tested this with good results (overall)?

If we use colorpicker (i think it's the best way), i would prefer arrays
json schema example: https://github.com/hyperion-project/hyperion.ng/blob/master/effects/schema/candle.schema.json#L20-L33

@redPanther what do you think? Any concerns?

Also related https://github.com/hyperion-project/hyperion.ng/issues/276

redPanther:
I also vote for arrays and color picker.
can we change config to something more compact and easier version
```
{
"red" : [255,0,0],
"green" : [0,255,0],
"blue" : [0,0,255],
"black" : [0,0,0],
"white" : [255,255,255],
"cyan" : [0,255,255],
"magenta" : [255,0,255],
"yellow" : [255,255,0]
}
```

penfold42:
Now if we could visualise it or find a cubic 3D colour picker like https://i-msdn.sec.s-msft.com/dynimg/IC726000.png it would be cool

brindosch:
Hehe http://codepen.io/pierre-luc/full/GqyxqG
But the handling is slighty fiddling. i think our current is a good start. You configure with a assistant each part step by step to prevent a 'overextension' the first time.

I wonder if it is possible to create something like a osd overlay (color picture) in current video stream for calibration. (a overlay of the current content) @redPanther is this possible?

Funatiq:
Yes, the array versions looks much cleaner. Color picker is nice, but I always like to fine tune the RGB values myself. 3D color picker is not very handy and I would prefer a flat one. A 3D visualization of the modified color cube would be cool though :)

I would like to keep the "black corner" and rather remove the blacklevel (and whitelevel) from the transform if we want to merge them anyway.

I only tested if the changes work with hypersim. I will do tests with my actual LEDs when I find some time.

brindosch:
> I would like to keep the "black corner" and rather remove the blacklevel (and whitelevel) from the transform if we want to merge them anyway.

Sounds good. Yes, i think the merge is necessary, cause you are going crazy to configure multiple calibrations with 2 arrays each. Rename and break whatever is suitable for a clean(er) code and removing task duplicates (whitelvl, ...).
> but I always like to fine tune the RGB values 

Our current one fulfill this. Edit the values -> cp updated, change cp -> update values.
unbenannt
More over this could be done - later (Or if someone is motivated - now) :)

Looking forward for more :)

redPanther:
If you can merge the color stuff and tidy up we would be very happy! Did you want to do this in this pr, or should we merge and you make a new one?

Funatiq:
I can update this PR, but it may take some time.

redPanther:
a suggestion: update at first only the config thing, this shouldn't be a big thing.
 Then we merge. After that we can think about merging transform and adjustment.

Funatiq:
Sounds good to me.

brindosch:
fixed typo...

redPanther:
Cool thing :-)

oberstmueller:
Hi, wenn ich bei der neuen Color Adjustment 2.0 die Werte ändere kann ich nur bei den Werten Schwarz und Weiß eine Veränderung feststellen. Bei den Farben ändert sich nichts.

Trotz meiner veränderten Config sehen rot, grün und blau unverändert aus
                                    {
                                                  "id"   : "default",
                                                  "leds" : "*",
                                                  "black"   : [0,0,0],
                                                  "white"   : [255,255,255],
                                                  "red"     : [255,30,0],
                                                  "green"   : [60,255,0],
                                                  "blue"    : [30,0,255],
                                                  "cyan"    : [0,255,255],
                                                  "magenta" : [255,0,255],
                                                  "yellow"  : [255,255,0]
                                    }

Funatiq:
Sorry, mein Fehler. Hab beim umbenennen was vergessen.
Hier der fix #357

oberstmueller:
Super vielen Dank für die extrem schnelle Hilfe. Jetzt passt es wieder!

brindosch:
@Funatiq 
Vorschläge für max gamma? Das steht gerade auf 100,0. Ich dachte da an 3,0. Ich hab auch noch nie jemand gesehen, der ein gamma von mehr als 2,9 hatte.

redPanther:
5 should be a good value

Funatiq:
5 sounds good to me and should definitely be enough

PR #358, opened 2017-01-03
Split initialEffect, rgb wizard, others

brindosch:
 - New: I wrote a wizard, "RGB Byte Order Wizard". Leave some feedback below this PR. Wizards are accessible via magic icon at the top.
 - New: New option in general to hide option explanations (for a better floating it needs a rework for all pages)
 - New: Enum lists from schemas are now translateable (and translated)
 - Reworked: InitialEffect is gone, welcome foregroundEffect and backgroundEffect. Both options are available at the effects page. (saving is disabled, there is one remaining issue with a not working change event trigger which results in an empty effect property until you change it once by hand)
 - Changed: Some effects use now integer values instead float to use a unit append (%). Delete your created effects, if they are affected.
 - Minor fixes and improvements, added partly titles for color area.

BREAKING CHANGES - compare with latest default


redPanther:
Wow great changes! I will test today or tomorow.
One thing I'm not sure if its good. I would rename foreground effect to initialeffect or boot effect, because this is seen on hyperion start. What you think?
Moving background efx to its own section is a good thing imho.

brindosch:
Possible, but a lot of renaming :)

Issue #338, opened 2016-12-17
HSL luminancegain is buggy


penfold42:
I have a large xmas led display with 600+ LEDs and I need to reduce the overall brightness.
With hsv  "valueGain"    : 0.1000 it worked great and continues to work great with #336 

With hsl "luminanceGain"    : 0.1000, it doesnt work well 8-(

If you set colours at the higher end of the brightness range - greater than about f8f8f8 it starts getting the resulting colour data wrong. 
1st column is requested colour, 2nd is the resulting led data sent:
```
ffffff	191919	perfectly fine
fffffe	323200	blue gone, red + green are double
fffeff	320032	same issue
feffff	003232	same issue
fefeff	000032	red and green missing
	
fefefe	191919	fine
fefefd	212110	blue is half gone, red and green to high.
```

am i missing something about the HSL colour space here?

redPanther:
It looks like precision/rounding problems

redPanther:
BTW luminanceGain 0.1 ??? My lights are nearly off. First value that is ok is 0.4.

redPanther:
can anybody try my branch:
https://github.com/redPanther/hyperion.ng/tree/hsltest

I've took conversion methods from:
https://www.programmingalgorithms.com/algorithm/hsl-to-rgb?lang=C
https://www.programmingalgorithms.com/algorithm/rgb-to-hsl?lang=C

perhaps that helps ;-)

penfold42:
Will try it today

0.1 is fine if you have a 750 led Xmas display and don't want to blind passing drivers !

redPanther:
> blind passing drivers 

ok, I guess the strobe effect is also not perfect for passing drivers :-)

penfold42:
I think I could dazzle planes at 3000 ft

redPanther:
or santa when he's comes by. ok now i understand the sense of extensive xmas lights. so long, gn8

penfold42:
At a quick look it is still broken.

Im wondering whether hsl scaling will ever work - it's quite possible the different co-ordinate systems need infinite precision at the top and bottom of the HSL shape.

redPanther:
I changed the precision from float  to double. Can you try again?

If this not working we can made lookup tables for the top and bottom values ....

redPanther:
Instead of  a threshold we can convert to hsv when threshold is reached. Transformation is easy ...
http://codeitdown.com/hsl-hsb-hsv-color/
At the ond of the page

[...]

PR #359, opened 2017-01-04
Adjustment merge + new brightness settings

redPanther:
- merge gamma from color transform to color adjust
- create new methods for brightness adjustments (aka backlight and upper brightness limit) in color adjust
- when source "off" selected, leds are off, regardless of minimal brightness
- full json support for new adjustment settings introduced by @Funatiq 
- remove colorTransform completly (and hope we don't have to bring it back)

**this pr breaks compat with prior config.** Please see details in default config.
Just in short: color transform section is removed and new settings for gamma and brightness are added.
Also hyperion-remote options and capabilities has chenged - cconsider this if you use it in custom scripts!

#338 #59 


Funatiq:
Looking really good. Thank you for the effort. I don't have much time to test atm. Did you test the backlight? A problem with the old backlight was that it had too much color, so we changed it to always be gray (r=g=b).
Additionally, maybe you could add a brightness multiplier to increase overall brightness? I don't know if we really need this with all the other changes, but I think a lot of users are used to the value/luminace gain.

brindosch:
Great changes!
Could you auto-adjust the brightness if (brightness <= minBrightness). As Hyperion quits currently, a log warning should be enough. How behaves this if you set a invalid brightness via jsonrpc?

redPanther:
now invalids brighness values leads to warning with no exception

brighness curve is now exponential. optimal setting is 0.5 (the mid), <0.5 it goes darker, but with equal brighness through the channels. > 0.5 it goes brighter, but channels are now not at equal brighness.

brindosch:
Ahh nice, will let you know how it performs.

redPanther:
the value curve:
hyperion_brightness
horizontal input value from config, vertical resulting sum(r+g+b) threshold

redPanther:
Backlight was the motivation, because old method annoyed me. For me its perfect now.

I think we dont need yet another brightness param. We have brightness nd white adjust, this should fit all needs.

Funatiq:
I agree on keeping the amount of parameters low.
What I don't get is why you set 0.5 as default. With this _sumBrightnessHigh defaults to 255, right? So everything with r+g+b > 255 gets darker per default. Wouldn't it make more sense to keep 1.0 as default to get the full color spectrum?

brindosch:
Well, the default could be changed. We  need more  feedback if a linear brightness (0,5) is better  than 1.0.    (which is on white 3x brighter than r/g/b).

Funatiq:
You are right when you think about the 3 different leds. My problem with this approach is that many colors above the threshold are the same.
Example:

image color | output color | my suggestion
---|---|---
255,255,0 | 127,127,0 |  127,127,0 
200,200,0 | 127,127,0 |  100,100,0 
150,150,0 | 127,127,0 |  75,75,0 

If we want to use this approach, the easiest solution would be to scale down the color calibrations.
black, red, green, blue -> no problem
cyan, magenta, yellow -> divide by 2 ~> something like (127,127,0)
white -> divide by 3 ~> something like (85,85,85)
And the interpolation would take care of all colors in between.

brindosch:
So i tested a little bit.
I notice the lost color values if i compare it. Your proposal is more accurate @Funatiq. This needs to be implemented in a way the user won't notice it (white: 255/(3*Linearfaktor, 255/(3*Linearfaktor), 255/(3*Linearfaktor)

On top i want to suggest to split maxBrightness and linearization. Which means you could control the brightness without the lost of linearization (you could also apply a different linearization to different brightness values)
@redPanther slack msg

oberstmueller:
Hi,
will there be no Saturation Setting in the new hyperion?

redPanther:
I hope we havn't bring it back. We want to simplify settings and save cpu time. Currently we tune color adjustment a bit.

For what you need saturation?

redPanther:
I hope we havn't bring it back. We want to simplify settings and save cpu time. Currently we tune color adjustment a bit.

For what you need saturation?

oberstmueller:
In some Scenes the Color is to strong for me. For example brown is oversaturated and looks like red.

PR #363, opened 2017-01-13
move write config from json api to http post 

redPanther:
- use http post to transmit the config from webui to hyperion
- update webserver code
- remove obsolete code for multicoloradjustment

now saving config should work as expected. A todo is the connection lost detection, this can lead to false connection lost messages (e.g. when network/device is slow)


brindosch:
You mean the reload of the ui?
You loose more than 2 packages in 4 trys?

redPanther:
Currently it checks if serverinfo could retrieved in 8 seconds for 3 trys.

My phone is sometimes slow, especially on loading. Once loaded ui is fast. I will move keep alive check to http. Websockets and retrieving serverinfo is to heavy

brindosch:
Hm, we should lower the amount of libs. I think jquery ui is nearly unused and could be replaced with css3 animations.

redPanther:
just found an issue (right after merge ...) spaces are replaced wit '+'. This is anoying in e.g. the back and foreground effect ... "rainbow+swirl+fast" instead of "rainbow swirl fast" will fix it asap

redPanther:
Now webui is stable on my slow phone. Besides webui is quite fast and also led/video preview works

PR #365, opened 2017-01-15
more stable connection lost detection

redPanther:
..


brindosch:
feedback: Sadly it's not failsave. This could be a restart of the service which is not introduced from the another client or even just a entire reboot as the fail of websocket connection doesn't matter - it hangs forever

brindosch:
With next update:
watchdog set to 1 and trigger conlost at websocket.onclose for an instant quit.

Issue #360, opened 2017-01-05
Limit priority range, priority recommendations


brindosch:
Strict:
I suggest to limit the priority range from 0-997 (3 digits are enough and don't eat much space)
Where 0 is reserverd for foreground effect
Where 1 is reserved for webui

json rpc should check against this range, before apply it. For the webui blocking we could filter against "origin" :webui (when it's implemented :))

background-effect could be moved to 998
off to 999
Clear-all. Apply it for 1-997? As we could keep the background effect running.

Recommendations
Remotes 2-99 (50 as default, they could overwrite each other)
Capture devices (100-997)

penfold42:
I don't think we should force anything.

Sensible defaults with good documentation explaining the convention should suffice.

I need to be able to override the webui - alexa Pi needs to be able to override pretty much everything for a few seconds

brindosch:
> I need to be able to override the webui
Well, okay. But 0 is forbidden :) (many custom scripts use 0, and i don't want that a application block the webui
The reason for a max priority of 997 are table layouts if you list all priorities and than a 2147483646 plops up. Could be a issue on applications due to low space and it looks overstated :)

redPanther:
we can limit the datatype of priority. currently it is int32 or int64 depending if we have 64 or 32 bit. We can change it to in16, than max prio is 32767

brindosch:
This is also something i thought first, but a even lower number with 3 digits would look quite good in an overview but it needs more code.
It's a start :)

yes, it's just eye candy - nothing fundamental

brindosch:
Okay, another crazy idea, to keep it simple.
uint8? It would be definitely enough, its just a matter of habit.

allocation:
platform capt. 250
usb capt. 240
boblight,udplistener 200
proto services 100-199
remote services 2-99
webui 1
foreground effect 0

Any concerns? I think it's better manageable (viewing)
And now don't argument you need more priorities than this!! ;D

redPanther:
looks good, but max is 255 not 256

brindosch:
of course! Edited
I will prepare the schemas.
It is not possible to change just one or some value/s to unit_8 right?

redPanther:
to get the of prio in range, change prioritymuxer.h:
`	const static int LOWEST_PRIORITY = std::numeric_limits<uint8_t>::max();`
if your changes complete, I can change the datastypes in the whole code

PR #366, opened 2017-01-16
priorities rework, webc update

brindosch:
- JSONRPC: moved activeEffects and activeColors to priorities
- JSONPRC/CONF: new priority limitations and guidelines: https://github.com/hyperion-project/hyperion.ng/issues/360 Update your configs!
- New: New settings level "advanced"
- New: lightning fast restarts and a new "restart" page
- New: config import/export
- New: detect script disabled browsers or blocking extensions and react accordingly
- New: You could now save as often you want without a restart
- Changed: led visualization is now reachable at the top via TV-icon in a drag and resizable dialog which stores the last used width and height across sessions
- Changed: editor no longer polls local storage and small button improvements that "fix" some usage issues
- Changed: led layout creation has 2 different "settings level" and all values are now stored/restored at local storage
- Changed: reworked all config pages, the content is now fully js created and panels float properly if help is off
- Changed: source selection reworked
- Changed: saving of a new lang or settingslevel triggers a browser reload. all code removed that was intended for change during runtime
- Fixed: wizardactive check no longer clears random web priority, prio selection resets now to auto if started
- Removed: Jquery ui and source maps
- lot's of code improvements, spelling mistakes, mores icons, disabled buttons when you shouldn't click them...


redPanther:
wow, as you predict - i love the changes especially the led view.

here some thoughts:
- can we change default prio for udplistener to 190? because it's not good when boblight and udplistener have same prio
- the gap between the top right icons are to high for small screens. Now I get 2 rows of icons on my phone.
- the new prio overview/source selection is extremly nice, but sadly to width for small screens. On small screens only the buttons should be visible, not the additional info
- can we add github to support -> information resource?
- horizontal buttons (and other form elements) (e.g. effect configurator, language selection, ...) overlap when they wrapped in a new row. the bottom margin should be set to some small value > 0 (e.g. 3 or 5)
(you can simulate screen size with firefox: "hamburger-symbol"->Entwicklerwerkzeuge->bildschirmgröße)
- when show explanations are off also the intro texts

brindosch:
Glad you like it

- yes we can, i just kept what we had before
- yes. I thought to remove the last icon, and in case we get +1 again - merge the options
- you should say very very small screen :) 367px is the breakpoint  where you get a second row on the top and this is also the point where the colorpicker floats out of the frame, it's not just the table :) I think a hidden origin column should be enough to reach the buttons again. Which screen size are we talking? 
- Yes, i also thought about a "about" page with all lib ressources, build number and such
- yes i am aware of the issue and i check regularly with different browser width how the floating behave. I thought often about how to resolve it without starting a space waste but margin seems the only way to get rid of it. I wouldn't call it overlap, its no margin. But some buttons float out of their parent frame, if you mean this :) )
- possible

if (screenwidth < 350) body display:none :D

jokes apart, for the top i can squash 15px, as they are added as nth.lastChild :)

redPanther:
The specs and sysinfo apps say 480x800. Browser reports 320x533.

redPanther:
I merge this. We can do improvments in next pr's

PR #374, opened 2017-01-22
Updated adalight sketch to use the number of leds sent in the header

penfold42:
- NUM_LEDS becomes MAX_LEDS and any led data beyond that is ignored

Added a python udp listener and serial sender.
- Intended for windows users with a COM port based led strip.
- Works with hyperion


redPanther:
Yes! This changes I always want to do but other stuff was more important ... thx!

PR #383, opened 2017-02-05
add clock effect

7h30n3:
And another effect. Just my little contribution to make this project even more awesome.

### DIFF
--- a/effects/clock.json
+++ b/effects/clock.json
@@ -1,0 +1,16 @@
+{
+	"name" : "Clock",
+	"script" : "clock.py",
+	"args" :
+	{
+		"offset" : 0,
+
+		"hour-margin" : 2,
+		"minute-margin" : 1,
+		"second-margin" : 0,
+
+		"hour-color" : [255, 0, 0],
+		"minute-color" : [0, 255, 0],
+		"second-color" : [0, 0, 255]
+	}
+}
--- a/effects/clock.py
+++ b/effects/clock.py
@@ -1,0 +1,75 @@
+import hyperion
+import time
+from datetime import datetime
+
+def myRange(index, margin):
+	return [i % hyperion.ledCount for i in range(index-margin, index+margin+1)]
+
+""" Define some variables """
+sleepTime = 1
+markers = [0, 13, 25, 38]
+
+ledCount = hyperion.ledCount
+
+offset = hyperion.args.get('offset', 0)
+
+hourMargin = hyperion.args.get('hour-margin', 2)
+minuteMargin = hyperion.args.get('minute-margin', 1)
+secondMargin = hyperion.args.get('second-margin', 0)
+
+hourColor = hyperion.args.get('hour-color', (255,0,0))
+minuteColor = hyperion.args.get('minute-color', (0,255,0))
+secondColor = hyperion.args.get('second-color', (0,0,255))
+
+""" The effect loop """
+while not hyperion.abort():
+
+	""" The algorithm to calculate the change in color """
+	led_data = bytearray()
+
+	now = datetime.now()
+	h = now.hour
+	m = now.minute
+	s = now.second
+
+	led_hour = ((h*4 + h//3%2 + h//6) + offset) % ledCount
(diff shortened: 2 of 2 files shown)


redPanther:
thx! can you add a json schema  file. This is located in effects/schema. This file is needed for the effect configurator and json validation

brindosch:
//done

brindosch:
@7h30n3 
i noticed after some testing that the minutes are at the wrong position for me.
Has it something to do with 12h 24h format?
Actually 15.00 o'clock is shown has 15.25. 

Thanks for any help

7h30n3:
@brindosch 
 I have to admit, I programmed this effect long time ago, but today it did a couple of tests with the Hyperion Effect Dev Kit.

At first  my setting (in main.py of the dev kit):
`horizontal_led_num = 16`
`vertical_led_num = 9`
`first_led_offset_num = -17`
`leds_in_clockwise_direction = True`
`has_corner_leds = True`

So my (50)LEDs start in the bottom middle of the TV, however I have to define an offset of `25` inside the clock.json the see the time displayed the right way. So offset inside clock.json is different form the first_led_offset.
I will look deeper into it at the weekend. Maybe you can solve your/the problem in the meantime.

It would be very good, if there would be a function to get the offset configured in hyperion, something like `hyperion.ledCount`

redPanther:
When you design the effect as an image instead of writing to the led stripe directly,then point 0,0 is always top left.
I done this in the light-céock effect

7h30n3:
Considering that there is already an clock effect with less configuration effort maybe we just should delete this one here.

brindosch:
//fixed and made a cleanup, you had a lot of vars and steps which have been never applied. Also i added a option for the markers to define them from ui (or disable them with -1)

https://github.com/brindosch/hyperion.ngBeta/blob/webb/effects/clock.py

Currently the clock effect of redPanther doesn't need a specific configuration, but i don't like the gradients. I wanted to add here anyway some options. Maybe one with and one without. And free editable colors.

I like the markers, could be also a good idea for redPanthers effect

7h30n3:
@brindosch 
I have seen the changes you made at:  https://github.com/brindosch/hyperion.ngBeta/blob/webb/effects/clock.py

Looks much better now. For `markers` maybe you should consider something like this:
```
marker = hyperion.ledCount/4
markers = (0, int(marker), (2*marker), int(3*marker))
```

So you can get rid of the option in the json-file.

brindosch:
Nice, as always don't forget the offset :)
I consider to keep it, but without a .json file. As everyone needs to adjust this. (But depends on how well i could adjust redPanthers clock)

So, one last question and as i have no idea of python.
I don't get a float here for the hour (at least it looks so), even forced with float(). I wanted to make it more precise.
Hints appreciated :)
https://github.com/brindosch/hyperion.ngBeta/blob/webb/effects/clock.py#L39

penfold42:
My clock runs backward...

Am i getting younger?

brindosch:
I envy you!

7h30n3:
@penfold42 Maybe you mounted your LED counter-clockwise. Maybe there should be a boolean to define the direction.

@brindosch I can't see any mistake in the function you mentioned. If I type this line in the python console (Python 3.6.0b3) and print the result I get a float value:
Let's assume its: 12:43

Start `cmd.exe`
```
python
>>> hmin = float(12+(1/60*43))
>>> print(hmin)
12.716666666666667
```
Same result without float().

You should also consider using `round()` inside/instead of `int()` (lines 45, 48, 51), so for example if you use:

-  Python 2: `int(round(12.716))`  becomes `13` and not `12`

- Python 3: `round(12.716)`  becomes `13` and not `12`

penfold42:
Correct - I would have fixed it myself but am zoo busy playing with some led device driver code

penfold42:
#395

brindosch:
Thank you for investigation, maybe a issue with py 2.7.
i think refer  to a inconsistent value is bad.

Am 10.02.2017 2:57 nachm. schrieb "penfold42" <notifications@github.com>:

PR #389, opened 2017-02-08
improve serial hotplug

redPanther:
- improve hotplug of usb serial device.
- add a pre connect delay. This is the time hyperion detects the device until it the open call. This is needed to give the OS a chance to set options to device before we block the device.
- add a basic auto detection for serial port name, just type "auto" as device name. But don't try this if you have more than one usb-serial device

additional:
- improve tpm2 firmware
- fix check json error 
- json checks now compat with utf8+bom
- as usual some cleanups

yes, if you use fadecandy. White point is now given as rgb color not as float values


libsrc/leddevice/schemas/schema-adalight.json
```diff
@@ -5,7 +5,7 @@
 		"output": {
 			"type": "string",
 			"title":"edt_dev_spec_outputPath_title",
-			"default":"/dev/ttyUSB0",
+			"default":"/dev/ttyACM0",
```
brindosch: I changed this due to the point it is ubuntu/debian convention. A special reason why you change it back? Did i missed something?

PR #390, opened 2017-02-08
Create cinamastart.json and cinamestart.py

IIIdefconIII:
A nice effect when the screen is refreshing to the movie, timings are perfect for several tv's i have. its an edit of the official system shutdown and without the yellow blincking on the end

### DIFF
--- a/effects/cinamastart.json
+++ b/effects/cinamastart.json
@@ -1,0 +1,10 @@
+{
+	"name"   : "cinemastart",
+	"script" : "cinemastart.py",
+	"args"   :
+	{
+		"speed"            : 1.2,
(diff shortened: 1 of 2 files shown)


brindosch:
Thank you for the contribution. In this case i would advice to add a bool option to the original shutdown effect which does the trick/skips it.

What do you think?

IIIdefconIII:
that would be awesome, but im not a dev and have no idea how to change this, i did just removed some lines until i got it working xD and changes the timing number

redPanther:
I don't like code dups, tahts why I close this pr. I added options to shutdown effect in my outstanding pr.

PR #397, opened 2017-02-13
[untested] added Sk6822 led device support

penfold42:
I've added Sk6822 led device support but haven't tested with real LEDs

    "device": {
        "colorOrder": "rgb",
        "invert": false,
        "output": "/dev/spidev0.0",
        "rate": 2230000,
        "rewriteTime": 5000,
        "type": "sk6822spi"
    },

nope


redPanther:
thx a lot! I also can't test it, but it looks good.

penfold42:
I tested the debug output and it looks correct.

Issue #395, opened 2017-02-10
Persist led layout settings in config


penfold42:
Would be good to have.

1) browser persistence isn't great if I change devices

2) some effects could do with knowing some of this info (clockwise, gaps, height and width.
The could be reverse engineered from the config as it stands but it would be horrible code

Obviously non standard layouts would have to default to NA or -1 if they make no sense.

brindosch:
For the second point i won't count on it. For a proper eff development you need values you could count on.
Also the logic behind to make sure that they are set to non sense values and if they are set make sure that the led creation switches into a fallback.

I you ask me, it is not worth the work, but @redPanther should leave a mind here :)

redPanther:
effects should based on image. Then all orientation stuff solves automatically.
For the clock effect: I want to merge light-clock and clock. I want add a param "faces" then we can have different clock faces. currently "artistic" and "analog". Before merge, I have to add line drawing to effect backend. (I also want to add all basic primitives)

brindosch:
Yes the image thing, i didn't understood how this really works.
Merging is good.

/closed

redPanther:
some stuff about image data in our effect engine:

we have 2 image implementations. I would call it 
1. "raw python"
2. "qpainter buffer"

image in general: you have an image with e.g. 40x25 pixel. This is projected onto leds in the same way as we project an image from grabber onto the leds. A pixel consists of 3 byte (RGB).

1) The image is stored in an array with size width x height x 3 bytes (e.g. 40x25x3=3000). Then you operate on that array to set rgb pixels you wish. E.g. to make a snake that crawl around your screen you have to set the outer pixel of the image.
If you set a pixel in the center of the image, then you will see this only if you have leds there (normale only on matrix layout)
The dificult thing is, that the 2D image must be accessed in an 1D array. 
example image size: 2x2
image:
```
(0,0,0) (255,255,255)
(255,255,255) (0,0,0) 
```
array representation:
```
0,0,0,255,255,255,255,255,255,0,0,0
```

Normally you access a pixel by x,y values and get the rgb color. But with 1D array you have to transform your x|y position to position in image array.
``` 
r = imageArray[ width*y*3 + x*3 ]
g = imageArray[ width*y*3 + x*3 + 1 ]
b = imageArray[ width*y*3 + x*3 + 2 ]
```

see knight rider effect for an example

[...]

PR #394, opened 2017-02-10
fix/refactor backlight stuff

redPanther:
**1.** Tell us something about your changes.
- fix colors dont turn off when backlight 0 and black is set (#391)
- add option to use not colored backlight (set threshold to lower value as used for colored backlight for best results)
- fix colored backlight not colored on very low color values
- various code style tunings

(this pr is a preparation for more tunings on brightness settings)

**2.** If this changes affect the .conf file. Please provide the changed section
yes:
brighnessMin -> backlightThreshold
new option: backlightColored

**3.** Reference an issue (optional)
#391

Issue #370, opened 2017-01-19
webUI should remember led layout parameters


penfold42:
Maybe we store these in the config file just for the benefit of the webUI?

If I want to repeatedly reconfigure the leds, maybe changing the the gap or depth settings I need to repeatedly enter all of the settings

brindosch:
i implemented a browser based persist. I thought also placing it in config as the gui is than always in sync with your layout. 
Downside is the bloat and you need to upgrade the config - if - you remove or add a option.

@redPanther

brindosch:
the browser based persist should work now..

PR #398, opened 2017-02-14
ledConfig in config, report creation and upload

brindosch:
- New: uploading log+config+basic info to https://glot.io (report) Demo report: https://glot.io/snippets/en4xcjfeld
- New: ledConfig is now stored in config instead storage. Matrix layout is not included!
- Fixed: wiz cc kodi address wasn't read from storage
- Changed: update remote and wiz cc cause of https://github.com/hyperion-project/hyperion.ng/pull/394
- Changed: clock effect is now usable. The default .json is no longer provided as you need to configure it first. Pacman and his guys have now a adjustable margin
- Changed: I added all possible keys for framegrabber to default config. We need a proper grabber information to hide options. e.g. if x11 is available - all non possible options should be hidden @redPanther  

Some words to the report upload.
glot.io is an open source project which is used to run code snippets, but you could also upload plaintext. This is not ideal, but due to the point you could upload more than one file that are grouped under one link and it allows CORS it is near perfect.
--> You could run glot.io on your own server and you are free how the presentation of the content should be. So for example for the config you could use a json syntax check/validation (inline!). You could create own code highlighter for the log (red for error).
You could also edit files and save them online (if you need to adjust something for someone).
---> All in all this will be on my ToDo list, if @RickDB  found some time for pending todos :)

@redPanther 
I am sorry, but i added again some leds :) Looks much better than just 3 left/right.

New object
```
	"ledConfig" :
	{
		"top"	 	: 8,
		"bottom"	        : 8,
		"left"		: 5,
		"right"		: 5,
		"glength"    	: 0,
		"gpos"		: 0,
		"position"	: 0,
		"reverse"	       : false,
		"hdepth"	       : 8,
		"vdepth"     	: 5,
		"overlap"  	: 0,
		"edgegap"      : 0
	},
```
And update your framegrabber section

https://github.com/hyperion-project/hyperion.ng/issues/370
https://github.com/hyperion-project/hyperion.ng/issues/395


penfold42:
Excellent work!

redPanther:
We can handle grabber config like the ledDevices. We can have schema files for every grabber.

brindosch:
I also thought about it. But than you need to remove the "auto" option. Hyperion needs to edit this in default config before it starts.

You need a detection for Pi BCM, x86 platform, Weteks and for other amlogic?
In any way, you need to do some actions. Which is why i thought just do provide a list of compiled capture devices and write just a some ui code wich hides option in a sort of priority chain (dispamanx is over framgrabber, x11 is over framegrabber. etc.

--> No need for additional detection or workarounds on default config creation.
But it's up to you. I will complete the however you wanna do it. :)

redPanther:
How about givin special options a prefix like x11_ dispmanx_ ... then its clear where the options belong and we have not so much changes.

brindosch:
Possible but doesn't look good. As they are also mixed. (some of them needs more than one attribute)
I think a list of cimpiled grabbers in serverinfo is for the moment good enough? (And not too much work?)

redPanther:
... you will get the list :-)

brindosch:
They "fixed" the CORS, upload is broken :)
@RickDB 
Time for own dbs :)

brindosch:
OT
@RickDB 
Forum emergency :)

RickDB:
Fixed :)

PR #407, opened 2017-02-28
json-rpc - origin, ui update

brindosch:
- JSON-RPC: For color and effect you need to provide an "origin" property which describes the setter (your application). It is not implemented for effects, yet.  @redPanther ideas how to pass this key fast to the prio register are welcome.
- New: led layout textfield is now a json ace editor with schema validation. @tpmodding this should also resolve switching issues you faced some months back
- New: WordPress 4.7 native rest api (blog posts retrieve). It's dead, as this is still a secret :)
- New: remote page "duration" for color and effect. There is a hidden "send image to Hyperion", you need to remove the display:none property to make it visible. But Hyperion can't handle it properly.
- New: all input fields outside of jsoneditor scope get a own validation that corrects user input.
- Changed: Restructured dashboard info for multi instance usage
- Changed: platform capture properties are hidden according to available capture devices
- Changed: con lost and restart clears cronId and they have a max limit of trys. The cronId also created every 2 seconds a new reconnect interval, which ended up in a reconnect-interval-hell that blasts your browser with 2GB ram usage for this tab away (after some time).
- Fixed: mobile nav switching around break point 765px (on orientation change), hidden on scrolling and scrollToTop that locks usage a certain time sometimes.
- Fixed: foreground and background effects could be saved (still a hack, was not my intention - in the and it won't matter much)
- Removed: flags, linedtext, underscore
- Removed meta in translation files, poeditor doesn't support it that way it would make sense.
- As always slight improvements, lang and access option moved to a wrench dropdown, as i see the next feature which forces this change anyway. Bootstrap is now a self compiled version where i removed everything i don't need. colorpicker updated.


tpmodding:
i'm speechless...

redPanther:
Cool stuff,thx! I will check the effect origin thingy.

Issue #349, opened 2016-12-24
Enhance priorities info


brindosch:
- Remove "ActiveEffect" and "ActiveColor"
 - Add this information to a priority, when it is a effect or color
 - For a better overview add a "who set it" identifier (App, WebUI, System (for capture devices) which needs to be added to your request (eg color, effect and proto sources)

An example
```
origin : AndroidApp
type : color
value : color values, duration
priority : 200
ip: 192.168.0.20

origin : Webinterface
type : effect
value : effect name, effect values, duration
priority : 250
ip: 192.168.0.22

origin : System
type : dispmanx
priority : 800

origin : HyperionScreenCap
type : proto
priority : 900
ip : 192.168.0.20
```

 - 'value' is just used for effects and colors with additional inormations that has been shown in activeEffect, activeColor
 - Duration is moved to 'value' due to the point a capture has no duration.
 - 'origin' is a identifier which is delivered from a request or 'system' if it is nothing external. requests with a missing identifier are rejected
 -  'type' replaces the content of the current origin, if it's a effect, the field contains 'effect' and the same for color with the values.

So leave some feedback :)
With this change it's perfect :)
@redPanther @Paulchen-Panther

penfold42:
Source ip address ?

brindosch:
Possible, but i think it's not that handy if you could already place a name instead (to remember). If you use the android app on a tablet and smartphone you could confirgure the origin name to your needs.

Or you need a ip for something different than identify for yourself?

penfold42:
Mostly for troubleshooting user problems on the forums. 

Things are going to get more complicated with home automation - nest, amazon echo, homegrown perl/python scripts etc...

brindosch:
update example
@redPanther what do you think? If this is done, it's perfect for(almost)ever? Maybe after the color area is finished :)

redPanther:
Yes, after color and webui saving

PR #408, opened 2017-03-01
implement origin for effects 

redPanther:
- implement "origin" for effects
- fix/add origin for network services
- fix log buffer QT signaling
- add experimental adalight  firmware for arduino mega (with upto 5 rgb pwm channles + 1 addressable led stripe)
- more conversion of STL types to QT types (mainly std::string to QString)

no

#349 

### DIFF
--- a/assets/firmware/arduino/adalight_mega_pwm/hyperion_mega.ino
+++ b/assets/firmware/arduino/adalight_mega_pwm/hyperion_mega.ino
@@ -1,0 +1,250 @@
+ 
+#include "FastLED.h"
+
+#define ANALOG_MODE_AVERAGE  0
+#define ANALOG_MODE_LAST_LED 1
+
+/**************************************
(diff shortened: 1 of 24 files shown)


brindosch:
protocomp - yeah!

socket->peerAddress()
Do we get always the current information of the peer who requested it?
--> Pass IP to prio register too :+1:

Very nice

redPanther:
yes we always get the client ip - otherwise we can't communicate. If you have a client socket, then you always have ip, mac and via dns the hostname of your connection client. (you get the client port to, but this is in most cases not interesting)

brindosch:
Amazing!
/prio improve done

redPanther:
BTW there is a chance that we automaticly support IPv6. I already get ipv6 as address for localhost (that's why I add dns lookup - ipv6 is not nice to display, read and remember ;-) )

Issue #411, opened 2017-03-02
migrate to qt datatypes


redPanther:
Hyperion is heavily qt based. Since we switched all external libs to qt it is senseful to migrate data types too.
Mainly it's std::string to qstring.
This can save unnessesary conversions and cleanup the code.

PR #412, opened 2017-03-03
migrate std::string to qstring + add sysinfo via json

redPanther:
- migrate std::string to qstring where it is usefull
- add sysinfo via json (hyperion-remote -s)

#411 


brindosch:
Amazing! Will use it :)

These lines could be removed
https://github.com/redPanther/hyperion.ng/blob/7ca0e65178aa59e6bc4afcc2720f62ea947bcae5/libsrc/jsonserver/JsonClientConnection.cpp#L596-L597

redPanther:
I thought about splitting serverinfo in a static and dynamic part.
'sysinfo' is catched on webui launch and contains stuff that doesnt change and 'serverinfo' contains stuff that changes and this is called continously

brindosch:
Q: How often you change a hostname, and bothers it that you need a restart of hyperiond to reflect it everywhere?

It's up to you, but i won't check hostname change on every poll on ui. (And i am confident other app devs won't do it too)

redPanther:
Its not what I meant. I want to move more things to sysinfo. E.g. the hyperion version numbers .... I will delete those lines you suggested ...

brindosch:
Ah, the entire static server info. Yes that makes sense
So add a checksum to serverinfo and push onChange to all open websockets with max interval of 500ms/1s? ;)

redPanther:
Now I integrated sysinfo call into webui. "sysinfo" is called only one time and "serverinfo" as before continuously.

PR #418, opened 2017-03-16
Reports - final version

brindosch:
 - Instead glot.io, we use now our own infrastructure. Reports are now available under report.hyperion-project.org. I call it v1 - as there is space for some nice functions.
 - To avoid upload dups, the last 5 report uploads are saved/shown
 - Removed some json parsing breaking output (In case there is much more i could also escape all...)
 - Some tiny refactor and optimizations, added more system information to reports.
 -  Added social links, these pages aren't done yet (content and banners).
- Added browser icons for any system
- entering invalid values for custom fields throws now a warning


redPanther:
fantastic!

Issue #353, opened 2016-12-28
toggle on/off multiple devices


tpmodding:
Hey guys, with the new webUI we can now do more things simpler...

can we add something that we can detect different instances by the config names and toggle them on/off or create scenes...
example:
scene 1:
ledLeft TV ledRight
scene 2:
         only TV

brindosch:
> detect different instances

First, you need this. https://github.com/hyperion-project/hyperion.ng/issues/55
> by the config names

The issues is now, you can't search for zeroconf information from your browser. @penfold42 do we need a population list via json rpc for webui to switch between?

redPanther:
Hyperion should search via zeroconf for other instances and populate it via json

redPanther:
@penfold42 can you implement (or give hints) a function that searches for other hyperions via zeroconf. 
Then we can use this to expose infos via json api

penfold42:
I probably won't have time for a while, but here is what influenced the existing advertise stuff:
https://doc.qt.io/archives/qq/qq23-bonjour.html#examplefortuneserverandclient

brindosch:
>time for a while

As time goes by
Looking forward for information sharing :)

Webserver config write is a huge issues for multi instances now. As you need a webserver for each instance and you need to assign them to the right json port!

redPanther:
I integrated bonjour browser and resolver:
https://github.com/redPanther/hyperion.ng/tree/bonjour
I bit difficult is that especially the resolver works async. Our json client is sync. Perhaps we have to make a async json call like:
1. webui query bonjour
2. webui gets "ok" for valid request
3. later on webui gets incomming json call for the requested list of hyperion services.

This methode is similar to our live streams for led/video preview

brindosch:
Closed
As of: https://github.com/hyperion-project/hyperion.ng/pull/578

PR #419, opened 2017-03-20
populate zeroconf/avahi/bonjour records via json api

redPanther:
search for other hyperion web ui's and populate addresses via json api (serverinfo).
Currently only webui is searched. Finding other json server is on the way.

additional:
- make leddevice as component - this can be en/disabled by json api/webui

... interesting part of serverinfo:
```
    "hyperion": {
        "config_modified": false,
        "config_writeable": true,
        "sessions": [
            {
                "domain": "local.",
                "host": "dory",
                "name": "Hyperion WebConfig@dory",
                "port": 8099,
                "type": "_hyperiond-http._tcp."
            }
        ]
    },

```

no

#353


brindosch:
Cool!
Could we work if ip's instead? At least i don't know how to blacklist the current connected session?
Not sure if we should rely on external dns lookup.

brindosch:
Uh, i waited for the compilation, but yes. I think you hacked it :)
{"active":true,"component":"Solid color","componentId":9,"origin":"System","owner":"Off","priority":255,"value":{"HEX":["0x000000"],"HSL":[0,0,0],"RGB":[0,0,0]},"visible":false}

I liked the id 0 for off. While it is actually rather unknown than assigned you could document: 0 = off. The componentId is the ONLY value which is consistent for filtering.
Well, i created another workaround for the missing origin of owner off. https://github.com/brindosch/hyperion.ngBeta/commit/8b8c499973d0cc62d43174a8533e04a627eb3405
+ removed component (duplicate and id is more effective), duration displayed just for color and effects.

redPanther:
We have to talk about what off means. Off could be:
- disable leddevice (perhaps adding leddevice to components
- disable leds and forwarder

Currently we send black to leds and forwarder. This have several disadvantages:
- changes the prio stack and we cant switch back to prior state when switch back to 'on'
- sends black to forwarder, every receiving hyperion session can't detect if sender is off or sends black color.

I think we should make the component thingy and off means disable leds and forwarder. When switch on again, set forwarder to prior state.

The current prio 255 is the background of prio stack and shouldn't be shown in json api prio list.
In webui we can make a on/off button. I suggest to rename the 'mapping mode box' to operation mode and place on/off there.

redPanther:
I like the hostnames instead of IP. hyperion can handle ip v6 and those addresses are mostly no fun. Other thing is, that behind one hostname can be multiple addresses ...
I splitted domain and hostname, so you can compare if you are on current session. You have to compare  session host+port and sysinfo host+current json port.

brindosch:
Hm, makes sense!
Now you could control it very individual.
Will check.

redPanther:
Whats missing is a "on/off" json command, but not in is pr .. will do it later ...

brindosch:
Works perfect, so fee free to remove https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/hyperion/Hyperion.cpp#L400

Another topic - the more i think about the compId the worse it is. If you remove now KodiCheck, the ids are broken upwards. Maybe it's better to refer always to componentToIdString and remove the Integer Ids from output (all!)

redPanther:
yes this point I forgott to mention - never use enums as number. in c++ use names and in js code use the strings. I will add it asap

brindosch:
And remove the -1 :)
https://github.com/hyperion-project/hyperion.ng/blob/master/src/hyperiond/hyperiond.cpp#L213

This might help to 'fix' clearall too?

brindosch:
Shall i do the cleanup?

redPanther:
you can implement webui stuff for "sessions". I will refactor on/off via json and do some cleanup in that process.

redPanther:
... keep in mind to use the "type" field. I will add other zeroconf stuff theire ... e.g. json stuff.

brindosch:
Proposal: Remove sessions that are gone. As they currently hang around there until you restart.

brindosch:
Hm, @redPanther
i think the idea with hostname is good. But you can never ensure that the router resolves it properly. I have a lot of hostnames in my network, but only one of them is properly resolved.
So i would recommend to switch back to ipv4, because it's simple and solid.

redPanther:
Problem is the zeroconf gives me the hostname and I have to ask dns for corresponding ip. When you look into the bonjour resolver I got a qhostinfo object. Sadly the address field is not filled.
Other thing is that the list cleanup is already implemented. I'll look into ....

PR #421, opened 2017-03-22
zeroconf updates

redPanther:
- add ip
- make dns service names more uniq (important for multiple local instances)

```
    "hyperion": {
        "config_modified": false,
        "config_writeable": true,
        "sessions": [
            {
                "address": "192.168.122.1",
                "domain": "local.",
                "host": "dory",
                "name": "MyHyperionConfig@dory:8098",
                "port": 8098,
                "type": "_hyperiond-http._tcp."
            },
            {
                "address": "192.168.122.1",
                "domain": "local.",
                "host": "dory",
                "name": "MyPi@dory:8099",
                "port": 8099,
                "type": "_hyperiond-http._tcp."
            }
        ]
    },

```

### DIFF
--- a/include/bonjour/bonjourrecord.h
+++ b/include/bonjour/bonjourrecord.h
@@ -55,6 +55,7 @@ public:
 	QString registeredType;
 	QString replyDomain;
 	QString hostName;
+	QString address;
 	int     port;
 
 	bool operator==(const BonjourRecord &other) const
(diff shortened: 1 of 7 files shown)


brindosch:
Much better, do you want to add the user friendly instance name for webconfig too? (So i am not forced to seach them at another session entry to display them)
> generalConfig["name"].toString("")

brindosch:
perfect!

redPanther:
Auto port is not so easy. For that we should modify  config to have a port range for each service, to avoid automaticly jam your network ...

Now we can modify the standalone tools to use zeroconf. Then a standalone grabber app finds hyperiond automaticly.

I will make a true on/off switch via json next.

brindosch:
As far as i know you get a response, if a port is already in use.
increment json +2
increment proto +2
increment webserver +1
If increment happend write it to config (So just one time (usually) to auto correct them).
Proceed as before :)

While zeroconf for standalone capture is good, i would rank the update for ports adjustment and start new hyperions from forwarder config much higher. :)

PR #423, opened 2017-03-23
Hyperion Switcher + cleanup

brindosch:
- JSON-RPC: property origin minLength 4 maxLength 20, removed title and component integer from serverinfo-components, min value for gamma adjustment is 0.1!, just to notice: @redPanther removed component from priorities and componentId ist now a string instead integer.
- Added switcher to top bar (just visible if there is something to switch)
- hidden service menu on about page (force restart and cache reset)
- better remote page floating on <1200px width
- clearAll for remote (requires fix for bg effect)
- prio 254 is ffa again

All concerns to me!

Migrate blackborderdetector threshold from float to integer e.g. `0.05` to `5` (see config file)


redPanther:
Ready for merge?

tpmodding:
So many updates...what a piece of luck that we have the webUI :D cant imagine to explane every new user how to do everything with the json file and text editor :)))

brindosch:
> piece of luck

I need to correct you, it's `piece of WORK`
:)

PR #424, opened 2017-03-23
global on/off

redPanther:
- hyperion-remote --off or --on to switch all components off/restore state before off

additional:
- set min gamma to 0.1

note fro webui: when we a toggle for on/off exists in remote, the component switches should be grayed/disabled.
When switched of, then component onoff is locked


brindosch:
ready for merge? I will add your requested changes

b1rdhous3:
We should a big red button on the home screen on webui :P

brindosch:
b!
Yes i created a "Smart Access" area at the dashboard. I think it's the best place. The status is also shown at the tom of infobox.

Issue #352, opened 2016-12-27
Expose json interface via python ?


penfold42:
Is this a good idea ?

I want to disable smoothing from some effects and this approach might have some other useful applications

brindosch:
Interesting idea,
i already thought how to get rid of smoothing in a good way. For some effects it's useful, others not so great (strobe for example). [Sometimes it could be slower or faster - but just for _this_ effect.]

penfold42:
If no one can think of another use for the json interface, we could add a smoothing option to the effects schema

redPanther:
Set smoothing on/off from within an effect is no prob, we just have to add a function to access hyperion setComponentState.
Changing smooth args is more work, because the args currently not changable on runtime.
I suggest to add on/off and dont care about the args yet'

redPanther:
smoothing is exposed to effects now. I wouldn't expose everything to effects. If some specific command is needed we can add it. (=> new issue)
I think we can close this issue.

brindosch:
Yes and if you push your idea to the next lvl @penfold42 you end with a "Plugin Python Engine". Which might be interesting :)
But this should be developed splitted from the effect engine anyway (looking to kodi C python api for missing inspiration :))

penfold42:
Agreed but it might be better to simply write a python (and other languages) library to make external python scripts easier rather than extend the hyperion core python

PR #425, opened 2017-03-25
disable smoothing for effects

redPanther:
- smoothing has no impact on effects. 
- merge strobe and fade effect
- add new breath effect
- prio 255 is black again, background color/effect is on 254 - just for safetiness
- fix pacman effect

#352


brindosch:
Cool!
prio 255 is just *reserved* and not used.
You need to revert the schema-color and schema-effect for allowed priority range

PR #427, opened 2017-03-30
Added brightnessFactor configuration option to LedDevicePhilipsHue

ntim:
Upon request from the forums, I added a new feature to the LedDevicePhilipsHue which enables the user to let the calculated brightness of the color to set of the Hue lamps be multiplied by an arbitrary factor. A factor of 1.0 (default) corresponds to no change. Factors > 1.0 result in a brighter output and thus more saturated colors. 
At the same time I refactored much of the device code which makes it more readable and more robust.


brindosch:
Cool, robust sounds always good!
I see you already detected the logger class :)
Could you add the new option to the json schema, this would be great! If you need help just ask.
https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/leddevice/schemas/schema-philipshue.json

ntim:
Done 👍

redPanther:
Big thx!

brindosch:
Thank you.
I noticed the public can't read reviews
I think this is a typo : `brightnessFactor = (float) deviceConfig["transitiontime"].toDouble(1.0);`
So it should be `brightnessFactor = (float) deviceConfig["brightnessFactor"].toDouble(1.0);`

Will update

brindosch:
I need further guidance
Hyperion segfaults with latest master with philipshue
```
hyperiond[559]: *** Error in `/usr/bin/hyperiond': malloc(): memory corruption (fast): 0x00d4b432 ***
Apr 03 18:12:46 raspberrypi systemd[1]: hyperion.service: main process exited, code=killed, status=6/ABRT
```
@ntim 
Could you also reduce a little bit the debug output. It's a little bit much :)

please verify @b1rdhous3 this issue

@redPanther 
If you may have an idea, just let us know :)

ntim:
Mmh will look into it this weekend...

ntim:
In the meantime, could you try to produce a stack trace with gdb?

b1rdhous3:
@brindosch what do I need to do to reproduce this issue?

brindosch:
I have one id with one led. Nothing special. I don't get a single short blink at the bulb while it seems to run shortly (hyperion)

brindosch:
@ntim 
As a first excerpt. I wonder why it fires so much get requests on startup, shouldn't it wait for response instead?

Why does it bother about the modelID? And why turns it the id on true/false again after that?

https://pastebin.com/0MczAtjM
I will check if i could get a gdb!

Thank you for you investigation.

brindosch:
I have some questions in general

- Do you need a output hz limit (e.g. smoothing to a specific output rate to reduce it)?
- Smoothing required for smoothing or could it be disabled (or set to 0 if above is true)?
- Do you have hscan/vscan recommendations for top/bottom/left/right "placed" bulbs? I used now values just out of my mind.

I want to provide perfect settings along with my new hue wizard (cool!)
Thank you for your help.

ntim:
It seems not all commits have been merged, will do a pull of the current master and see whats left.

ntim:
Ah ok, github is a bit weird here stating my fork is 8 commits ahead of the master

ntim:
To answer your questions:

* Do you need a output hz limit (e.g. smoothing to a specific output rate to reduce it)?
There is a limit of about 30 instructions per second for the hue system. Thus, the framerate of the grabber should be around ~ 5 Hz for two bulbs to be on the safe side. 

* Smoothing required for smoothing or could it be disabled (or set to 0 if above is true)?
Smoothing does not make sense to me since it increases the refresh rate of the bulbs

* Do you have hscan/vscan recommendations for top/bottom/left/right "placed" bulbs? I used now values just out of my mind.

For two bulbs (one left and one right of the TV set) I use

```
"leds":
	[
		{
			"index" : 0,
			"hscan" : { "minimum" : 0.0000, "maximum" : 0.0700 },
			"vscan" : { "minimum" : 0.0000, "maximum" : 1.0000 }
		},
		{
			"index" : 1,
			"hscan" : { "minimum" : 0.9300, "maximum" : 1.0000 },
			"vscan" : { "minimum" : 0.0000, "maximum" : 1.0000 }
		}
	]
```

brindosch:
My initial idea was to use smoothing to reduce the refreshrate data hits the led device. (without smoothing anything).

Bad
The solution is this: https://github.com/hyperion-project/hyperion.ng/pull/432
:)

To prevent issues or add further config tuning you might want to calculate (internal) dependent of light ids count your own latchtime

Am i right @redPanther ?
Thank you for your efforts @ntim 
Looking forward for your completion pr :)

ntim:
I do not follow. The code does not except in the old hyperion where we had one thread, right? Now as I see in gdb there are multiple threads accessing the leddevice at the same time but my code is far from thread save since there is a feature which restores the state of the bulb after a given time, i.e. hyperion is still running but nothing changes (pause, stop, etc.). 

So summing up, there is a problem when hyperion tries to write some new color values while the leds are turning off. Thus, a rate limiting is not the solution but a intermediate band aid in my opinion.

redPanther:
rate limitation is now merged. Feel free to tune min,max,default value of https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/leddevice/schemas/schema-philipshue.json#L48

BTW "0" for latchtime means rate limiting disabled

brindosch:
There wasn't any change @ntim . You should also notice that the previous code worked so far. If there is now a huge issues, it was just hidden?
A restore of the previous state makes indeed sense.

brindosch:
In case it is not fixable we sadly need to revert this pr.

ntim:
First, can someone answer my question regarding the multi threading? For me the new Hyperion version never seemed to run smoothly with hue but also could not really test it on osmc since qt5 does not work out of the box. This pr is just a copy of the code from my fork of the old hyperion in which it runs perfectly.

brindosch:
> since qt5 does not work out of the box - osmc

Hm? I use it daily. Install and you are fine? What issues do you have? https://github.com/hyperion-project/hyperion.ng/blob/master/CompileHowto.md

> multi threading?

I just know there wasn't a change but maybe the gdb log would be helpful that @redPanther could have a quick look to assist/inspect

ntim:
Ok, when I tried to install it, qt5 caused osmc to crash which was a known problem then. I will give it a try...

redPanther:
There was no change in threading. There is just the main thread and one thread per effect.

ntim:
So are there multiple instances of the led device created and / or should it be threadsafe?

redPanther:
No multiple instances! Led device is not thrwadsafe, but this is not needed atm.

redPanther:
There is a qtimer for the refresh interval. Perhaps that could lead to race conditions. Will add mutexes asap.

brindosch:
@ntim 
Any new results based on redPanthers hint?

brindosch:
@ntim

Issue #422, opened 2017-03-23
Adjustments bugs


brindosch:
For your next session @redPanther 
- Gamma of 0.0 results in white (raise min to 0.01?)
- maxBrightness of 0.0 results in 1.0 brightness (independet of minBright) (raise min to 0.01?). I vote for integer values anyway :)

redPanther:
gamma 0 -> white is not a bug, its just what gamma does. But yes we should limit the range. Mostly gamma < 1 makes no sense in our case. will set the min to 0.1

here the gamma crve for gamma: 0-1 and 2
![gamma_curve](https://cloud.githubusercontent.com/assets/6893049/24251096/4c2c36c0-0fd9-11e7-8a9d-193e17880d39.png)

PR #430, opened 2017-04-02
new brightness handling

redPanther:
this is a result of lot's of discussions. The new brightness stuff is based on @Funatiq new color correction. Setting the brightness modifies the values for color correction. This will limit the brightness of each color component. The new option "brightnessCompensation" handles the problem that e.g. white (255,255,255)has 3 times more energy than red (255,0,0).

additional: updatewebserver, now with initial ssl support

- brightness are now int/percent values from 0 to 100
- new value brightnessCompensation. This is also a percent value. 100 means all colors are equal bright, 0 the opposite

#422 


Funatiq:
I like what you did with the brightness. I have some questions:
- You initialize every color value separately for every input (_initialized[input]). Any reason for this instead of the complete initialization?
- Can you explain your idea here:
`B_in = -0.093*_brightness+std::sqrt(_brightness);`
`if (_brightness< 57) B_in += std::sqrt(57)-std::sqrt(_brightness);`
Is a simple linear function not sufficient?

Btw I would have preferred if you had separated the brightness changes and the webserver changes in different PRs. Would have been easier to read ;)

brindosch:
perfect so far, i didn't found weirdness. I wonder if you want to move the backlightThreshold to Integer too (for the visual touch :))

redPanther:
The maping of config value is not linear because it feels better. I tried that 50% is half brightness
I update the mapping val's just in time, because do it in one piece means 7*256 calculations. The request for new brightness can come fast and I dont want to make unneccessary calculations.

Funatiq:
> The mapping of config value is not linear because it feels better. I tried that 50% is half brightness

Ok, I understand. But sqrt is quite expensive. I would suggest a piecewise linear function like [this](https://www.wolframalpha.com/input/?i=Plot%5BPiecewise%5B%7B%7B-0.09*b+%2B7.5,+b+%3C+50%7D,+%7B-0.04*b+%2B5,+b+%3E+50%7D%7D%5D,+%7Bb,+0,+100%7D%5D). It's almost the same as [yours](https://www.wolframalpha.com/input/?i=Plot%5BPiecewise%5B%7B%7B-0.093*b+%2Bsqrt(57),+b+%3C+57%7D,+%7B-0.093*b+%2Bsqrt(b),+b+%3E+57%7D%7D%5D,+%7Bb,+0,+100%7D%5D).

redPanther:
thats a good idea, will modify the code tomorow

redPanther:
now backlightThreshold is 0-100
brightness uses piecewise linear func as suggested

just for interessed:
hyperion_brightness

this shows mapping of brightness value to max color value at brightnessCompensation=100

PR #432, opened 2017-04-07
add limit for write to leddevice

redPanther:
... writing to fast to a led device can led to problems.
For arduino/adalight is 20ms a good value. (my experience - otherwise I got flashes)

additional:
- extend xmas effect
- rework some effects to adapt to device latchTime
- fading for random.py

yes, new option "latchTime" to "device" section

### DIFF
--- a/effects/schema/x-mas.schema.json
+++ b/effects/schema/x-mas.schema.json
@@ -5,11 +5,47 @@
 	"required":true,
 	"properties":{
 		"sleepTime": {
-			"type": "number",
+			"type": "integer",
 			"title":"edt_eff_sleeptime_title",
-			"default": 1.0,
-			"minimum" : 0.1,
+			"default": 1000,
+			"minimum" : 100,
+			"append" : "edt_append_ms",
 			"propertyOrder" : 1
+		},
+		"length": {
+			"type": "integer",
+			"title":"edt_eff_length_title",
+			"default": 1,
+			"minimum" : 1,
+			"propertyOrder" : 2
+		},
+		"color1": {
+			"type": "array",
+			"title":"edt_eff_color_title",
+			"format":"colorpicker",
+			"default": [255,255,255],
+			"items" : {
+				"type": "integer",
+				"minimum": 0,
+				"maximum": 255
+			},
+			"minItems": 3,
+			"maxItems": 3,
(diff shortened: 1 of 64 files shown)


effects/schema/x-mas.schema.json
```diff
@@ -5,11 +5,47 @@
...
+			"format":"colorpicker",
+			"default": [255,0,0],
+			"items" : {
+				"type": "integer",
+				"minimum": 0,
+				"maximum": 255
+			},
+			"minItems": 3,
+			"maxItems": 3,
+			"propertyOrder" : 3
```
brindosch: propertyOrder of line 34

penfold42:
Is this the same as latchtime but done globally ?

Maybe we could combine them.

The time needed for store and forward devices like an arduino will need to tuned based on the number of leds

redPanther:
Its the same. I uèdate the code later

redPanther:
I want to remove latchtime. Do we need ns precision?
Btw the diff between latchtime and this implemention is, that latchtime blocks the whole hyperion, while the other just drops the frame.

penfold42:
Nah...  

Latchtimes from datasheets I had handy: 
Ws2801 - 500 uSec
Ws2812 - 50 uSec
Ws2813 - 300 uSec
Apa102 - none - it's sent as end frame data bytes

redPanther:
setting 1ms or perhaps better 5ms will satisfy all needs. To be flexible I will change the code, so every device can have its own default value for latchtime.

penfold42:
Even at 5ms we should be ok. At some point it will be a limiting factor on the frame rate.

200 leds * 24 bits = 4800 bits
At 1,000,000 bit/second = 4.8 ms
That's 208 fps 

Adding 5ms of default wait time is still > 100 fps

redPanther:
now old latchTime is gone and new latchTime is implemented. For the device specifc latchTime params (min, max, default) I need suggestions. Currently I guessed for most devices.

penfold42:
Does the latchtimer start after the write is started or completed ?

redPanther:
It starts before write. The check is also before write. If timeout is not reached, the device write is not called.
So the latchtime includes the data transfer.

PR #433, opened 2017-04-13
extend image based effect funcs

redPanther:

- add func for setting minimum size of effect buffer image. This must be done asap in the effect
```
# scale buffer minimum to 64x64 , keep aspect ratio
# you can get real size of buffer with hyperion.imageWidth() hyperion.imageHeight()
hyperion.imageMinSize(64,64)
```
- add posibility to save a image to memory list. Func returns the index  of the saved picture. You can pre draw your frames and then show just the images in the loop.

```
imgList = []
hyperion.imageSolidFill(255,0,0)
imgList.append( hyperion.imageSave() )

hyperion.imageSolidFill(0,255,0)
imgList.append( hyperion.imageSave() )

hyperion.imageSolidFill(0,0,255)
imgList.append( hyperion.imageSave() )

# Start the write data loop
while not hyperion.abort():
	for img in imgList:
		hyperion.imageShow(img)
		time.sleep(1)
		if hyperion.abort(): break
```

### DIFF
--- a/effects/light-clock.py
+++ b/effects/light-clock.py
@@ -2,8 +2,8 @@ import hyperion, time, datetime
 
 # Get the parameters
 showSeconds = bool(hyperion.args.get('show_seconds', True))
-centerX     = int(round(hyperion.imageWidth)/2)
-centerY     = int(round(float(hyperion.imageHeight)/2))
+centerX     = int(round(hyperion.imageWidth())/2)
+centerY     = int(round(float(hyperion.imageHeight())/2))
 
 colorsSecond = bytearray([
 	0, 255,255,0,255,
--- a/effects/rainbow-swirl.py
+++ b/effects/rainbow-swirl.py
@@ -1,5 +1,8 @@
 import hyperion, time, math
 
+# set minimum image size - must be done asap
+hyperion.imageMinSize(32,32)
+
 # Get the parameters
 rotationTime = float(hyperion.args.get('rotation-time', 3.0))
 reverse      = bool(hyperion.args.get('reverse', False))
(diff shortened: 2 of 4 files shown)


brindosch:
Cool!
Will test asap.
Will expand docu too.

Did i mention we got touch drag support for led visualization?
 https://github.com/atatanasov/gijgo/issues/48#issue-201351800
Thanks to @atatanasov

brindosch:
I tested hyperion.imageMinSize(64,64)
Hm, my intention was to have a hugher image playground, Actually it is a hugher playground, but zooming wasn't my thought, as you zoom it's just a part of a picture now visible!
I thought about a version without zooming. Like usual capture pictures -> led area calc. This should smooth all (diagonal) lines, gradients with hard color stop and so on.
Also: As you said, imageH/W can't be used with it, it looses center.

Request:
drawDot(), i like dots! They are so DOTish!
Don't be a dot, dot!

ImageSave, very cool!

redPanther:
After minsize is called the image buffer is resized. It keeps ratio with hardware leds.
You can use image w/h for calculate the center, because this is updated after resize.

It is important to call minsize before any drawing actions.

I had to use the scaled func of qimage because there is no other way to set new w/h without destroying the image object.

brindosch:
Ouh, wow the color calcs are somehow... different :D
I need to buy a led matrix!...
It works as intended. Thank you

redPanther:
You get the ...... dot, will do it in the next time slot. BTW now we have the image stack. Next we can build animation methods and use image filters (e.g. blur, wave, ripples, ....)

redPanther:
http://doc.qt.io/qt-5/qgraphicseffect.html

Issue #437, opened 2017-05-06
Need to change ws2811 reset timing 


penfold42:
Like here https://github.com/forkineye/ESPixelStick/issues/47

Our standard arduino sketch might similar fix

redPanther:
> Our standard arduino sketch might similar fix

fix is done with taking latest fastled ...

PR #439, opened 2017-05-13
Ammeded timing to suit new revision of ws2812c

penfold42:
Ammeded timing to suit new revision of ws2812c
Should also work with WS2813 (but untested)

Included comments on the timing calculations

nope

#437 

### DIFF
--- a/libsrc/leddevice/LedDeviceWs2812SPI.cpp
+++ b/libsrc/leddevice/LedDeviceWs2812SPI.cpp
@@ -1,8 +1,43 @@
 #include "LedDeviceWs2812SPI.h"
 
+/*
+From the data sheet:
+
+(TH+TL=1.25μs±600ns)
+
+T0H,	 0 code, high level time,	 0.40µs ±0.150ns
+T0L,	 0 code, low level time,	 0.85µs ±0.150ns
+T1H,	 1 code, high level time,	 0.80µs ±0.150ns
+T1L,	 1 code, low level time,	 0.45µs ±0.150ns
+WT,	 Wait for the processing time,	 NA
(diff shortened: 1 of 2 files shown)


redPanther:
btw there is a  new version of ws2813: https://github.com/FastLED/FastLED/issues/449
the older one is not compat with the newer one. I use the new one now with fastled 2812 driver

penfold42:
Oh goodie !

I think I'll write a generic SK-WS 1 wire led device that exposes all the timing as config options.

Until I do that, at least I provided detailed comments to enable quicker rework

redPanther:
With increase of ws2812 reset time to 300ns for 2017 produced 2812 and 2813 should have fairly same timing.

Btw did you notice the upcomming 2815 led? This is a 12v droven stripe. I hope this will become available soon, so we dont have to deal with high amps for larger lef installstions.

Another info is that 2812 is declared obsolete by worldsemi and 2813 will be the replacement

penfold42:
I hadn't heard either - thanks for sharing.

12v brilliant ! I hope they do 3v3 compatible signalling too.

brindosch:
Pictures looks promising, i hope it's not a fake and the lost power isn't crazy high (burned with resistors) :)
The 5V are really annonying. 
@redPanther Slack :)

redPanther:
http://www.world-semi.com/solution/list-4-1.html

redPanther:
https://blog.adafruit.com/2017/05/03/psa-the-ws2812b-rgb-led-has-been-revised-will-require-code-tweak/

penfold42:
Yep - that's the reason for this PR

i wonder if they're using the same die as the 2813

PR #443, opened 2017-06-12
Websocket auto serverinfo responder

brindosch:
 - Adds the fancyness of event driven serverinfo callbacks for websocket clients. Therefore the webui responses instant to changes. A polling of serverinfo is no longer required, which avoids network traffic and useless client calculations.
 - The push is instant, but the max interval is set to 250ms in case of repeated push requests.
 - Effects and colors with duration higher than 0 have a own 1s interval, as you usually display the time in seconds.
 - All websocket clients will be notified about the change, which keeps everything in sync.
 - **UNTESTED** are boblight and udplistener, they require for sure some own triggers to inform the hyperion class for changes, which affects the serverinfo.
 - Everything else included self quitting effects should work.
 - cut between serverinfo request and config check, fixed interval of 2 seconds. Maybe we should also cut priorityMuxer updates from the hyperion::update() and set it to a fixed interval.

Have fun
Ping @penfold42 , @Paulchen-Panther, @redPanther, @deespe

// added gif
serverinfo_cb


brindosch:
// added gif

brindosch:
No feedback is also good! Please note all outstanding issues here.

penfold42:
That's cool !

PR #444, opened 2017-06-21
Move JsonProcessing from JsonClientConnection to own class

brindosch:
- The Json-RPC processing is no longer part of the JsonServer(JsonClientConnection)
- As own JsonProcessor class it handles all data exchange (JsonForward, serverinfo notify, serverinfo trigger) directly with hyperion.
- JsonServer is now responsible for jsonforward instead JsonClientConnection
- Hyperion is now responsible to limit the serverinfo forced pushes instead JsonServer
- All these changes flattens the way for: HTTP JSON-RPC, UDP JSON-RPC, **insertAnyOtherNetworkService**

Usage
```
#include <utils/JsonProcessor.h>
// spawn instance
JsonProcess * jsonProcessor = new JsonProcessor(QString address);

// listen for callback Messages (QJsonObject), please note that you don't need to call handleMessage() as the forced serverinfo will also trigger them
connect(jsonProcessor, &JsonProcessor::callbackMessage, this, &MyClass::sendReply);

//process a command
jsonProcessor->handleMessage(QJsonObject)
```
That's it. input your commands from request and get the callback to send it as reply.

@penfold42 :)


penfold42:
Jason is a kid I went to school with.

Do you need me to update something ?

brindosch:
>Jason is a kid I went to school with.

Why not :)
It was more a FYI in case of expanding udp capabilities :)
However, the tcp implementation works very well so far (based on latest feedback from BioHazard1), so it's not _required_ but might be a option in future

PR #449, opened 2017-07-04
Add zeroconf TxtRecords

brindosch:
- unquie instance id + version in zeroconf txt record
- @penfold42 FYI - It can't be easier :)
unbenannt

@tavalin Your static id without messing with hostname and user firendly names for uuid generation.


tavalin:
Awesome, thanks @brindosch! Will pull the changes at the weekend and test.

penfold42:
Nicely done - I was looking at arduino mDNS a couple of days ago and it looked like a list of pairs. I suspect you can have multiple TXT records.

It doesn't look too  pretty but for our purposes it'll do.

Well done !

brindosch:
> It doesn't look too pretty but for our purposes it'll do.

Yes, could be better. I also thought about to just set it static for all instead passing around another argument. But that way you can pass whatever you want as much was you want. Not sure if we ever want to expand it or update these entries during runtime (also possible).
For example for http you can add the page entry url and such stuff. I keep it modular for now. 

@tavalin 
thank you, just report back if you have issues or strange stuff going on.
Did you found the issues with the plop up printers?! I really wonder, because the service string to scan for is very clear. If we need to adjust it, just let me know.

PR #453, opened 2017-07-30
hyperiond desktop integration

redPanther:

on desktop systems it can be nice to have hyperion as desktop app and not as system service. 
This PR adds optional desktop support.

In desktop mode:
a systray icon is shown. From there you can:
- open settings page in browser
- set color/effects (and clear them; prio 1, like webui remote)
- quit hyperiond

hints:
- osx is always in desktop mode
- on linux auto behaviour  is only available when compiled with ENABLE_X11
- this is only active if your desktop environment supports systray icons

In service/console mode:
- all things as before

When hyperiond is started it decides in which mode should be used, depending on whats available. You can override auto settings with commandline args: [code]--service and --desktop[/code]

To make it more easier for gui integration, now a default config file name is set, when no config file is given.
Normally your home dir is chosen. If this is not writeable then the dir where hyperiond is, is selected for config file.

All new things are optional! 

no


brindosch:
wb!
Well, i think the icon at /webconfig is slightly misplaced :)

I come back to you regarding readyRead and waitForReadyRead tomorrow :)

redPanther:
I've reworked my pr and made it a bit more complete

b1rdhous3:
It is working on osx.

Issue #447, opened 2017-06-26
Effects smoothing+


brindosch:
I want to bring back a point which isn't properly solved. Effects and smoothing. I see the follwing issues:
- Some effects require it (candle, mood blobs) and others should use it too, instead running on device latchtime border. Well, it is smooth, but the cpu usage of hyperiond is on a Pi2 at around 40%, for the swirl which is very high.
- Using the global smoothing for control about this is weird. The values might be too high and what happens if you run 2 effects which wants to enable and disable it? Check against the active one again?

From a code view it might be better to implement a own smoothing for the effect engine which could be build more dynamically for better control from effects.

https://github.com/hyperion-project/hyperion.ng/issues/352
@penfold42

penfold42:
Personally I'd rather add smoothing settings to the effect schema and enhance the existing smoothing to be dynamic rather than just off/on

redPanther:
I have a nice idea. I'm on it ... stay tuned

redPanther:
issue solved?

brindosch:
I think so. Thank you
Will later on tune everything.

PR #456, opened 2017-08-03
per effect smoothing

redPanther:
this pr adds ability to set per effect smooting settings.
Smoothing settings are written in effect json file.
As default all effects without smoothing params in json config have smoothing disabled.

Currently I added smoothing settings only to effect "candle" and "random".
(others should follow, but I don't feel in a artistic mood today. Feel free to post settings for a specific effect as comment to this pr.)

new candle.json
```
{
	"name" : "Candle",
	"script" : "candle.py",
	"args" :
	{
		"sleepTime" : 0.20,
		"brightness" : 100,
		"color" : [ 255, 138, 0 ],
		"candles" : "all",
		"smoothing-custom-settings" : true,
		"smoothing-time_ms" : 500,
		"smoothing-updateDelay" : 0,
		"smoothing-updateFrequency" : 20.0,
		"smoothing-pause" : false
	}
}
```

Additionaly:
- refactored some qt vs. stl stuff (mainly std::min/max to qMin/Max).
- fixed fade.py effect (crashed with div by zero)

#447


penfold42:
Perfect !

Great work

brindosch:
Very cool!
You can hide options by adding a dependency
```
			"options": {
				"dependencies": {
					"smoothing-custom-settings": true
				}
			},
```
So this option won't be shown until smoothing-custom-setting is true

redPanther:
cool I always wanted such option :-) it's integrated now

redPanther:
testbuild complete. from my side ready to merge. We can add custom smoothing settings to other effects later on ...

brindosch:
I have a question, is there a real diff between "enabled" and "pause" during usage? If you disable it, it is on kind of pause anway?

redPanther:
pause is internal for effects and enable is for the user. 
technically pause feeds the smooth device and enabled does it not.

We can rename 'pause' in 'enabled' in effects json ... this would look a bit nicer for users.

brindosch:
So it does simply the "same". I think a merge is useful to reduce size!

Issue #466, opened 2017-08-30
Philips Hue


brindosch:
Philips Hue is unstable in it's current state. Following a list of notes
- Check if bridge ip is reachable, verify that user has access. Try again after 10 seconds if failed
- Adapt update speed to the configured light count (maybe with speed selection slow/normal/fast)
- In case the bridge goes offline, step back to point 1
- Remove QEventloops, use async feedback. (thread freezer)
- switchOffOnBlack broken(?) verification required
- led count vs light id - untested

@ntim @bimsarck 
You are invited (also to test my webui assistant :))

seblagaffe:
Hy,
The webui is great. But on restart, after boot effect (which work) hyperiond crash 
`[hyperiond EFFECTENGINE] <INFO> effect finished
[hyperiond Core] <INFO> Unregister input source named 'Rainbow swirl fast' from priority register
[hyperiond LedDevice] <DEBUG> <LedDevicePhilipsHue.cpp:152:post()> Post http://192.168.0.11/api/PDfs6Ld0l0QxUaVTF6C2tzWoy9nivFMzmg0IcVUP/lights/1/state: {"on":false}
[hyperiond LedDevice] <DEBUG> <LedDevicePhilipsHue.cpp:152:post()> Post http://192.168.0.11/api/PDfs6Ld0l0QxUaVTF6C2tzWoy9nivFMzmg0IcVUP/lights/2/state: {"bri":56,"on":true,"xy":[0.33700000000000002,0.33160000000000001]}
[hyperiond LedDevice] <DEBUG> <LedDevicePhilipsHue.cpp:152:post()> Post http://192.168.0.11/api/PDfs6Ld0l0QxUaVTF6C2tzWoy9nivFMzmg0IcVUP/lights/3/state: {"bri":118,"on":true,"xy":[0.159,0.27160000000000001]}
[hyperiond LedDevice] <DEBUG> <LedDevicePhilipsHue.cpp:152:post()> Post http://192.168.0.11/api/PDfs6Ld0l0QxUaVTF6C2tzWoy9nivFMzmg0IcVUP/lights/1/state: { "xy": [0.0000, 0.0000], "bri": 1 }
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 1) >= this->size() (which is 0)
`

cmuckel:
Do you have a solution for the broken switchOffOnBlack? I noticed the problem yesterday

[...]

Issue #429, opened 2017-04-01
Not all Hue recognized


Gamer106:
I've gotten Hyperion setup, and the WebUI breaks generally anytime I try to save/restart (Raspbian Jessie Pi3), and when it does restart it'll enter a failed state either due to not recognizing the devices, or not enough bulbs (easily fixable by editing the number in Index). Is there something I need to do so that more of the lights are discovered or recognized? I have 6 total. 3 Bulbs, 1 Bloom, 1 Go, and 1 Lightstrip. Only the Bloom, Go, and one bulb are recognized/working.

BEFORE YOU OPEN A NEW ISSUE!

Search at our wiki: wiki.hyperion-project.org
Our FAQ: https://hyperion-project.org/wiki/FAQ-Frequently-Asked-Questions
And have a look at our forum: forum.hyperion-project.org
If you need help please open a new thread their!

WE PROVIDE NO SUPPORT AT GITHUB!
All misleading issues will be closed without a announcement!

b1rdhous3:
I can confirm this. After the newest PR's my hyperion starts for about 10 seconds and closes then. 
I think there is something broken with hue atm.

Gamer106:
Ah, understood. That's what's happening when it eventually does work. They'll do the intro, then I can't calibrate the colors because they stop :(

brindosch:
Please provide feedback at PR472, unrecognized ids are now properly logged.

PR #472, opened 2017-09-09
Philiphs Hue Improvments

brindosch:
This is a patched, rewriten and improved version of the philiphs hue implementation
Changes:
- segfaults segfault segfaults. No segfaults! Solid against wrong configuration, bridge not reachable, unknown light ids and so on
- No thread freezers anymore which leads to unexpected behaviour like closed sockets or not responding sockets of the hyperion api
- check if bridge is reachable, user has access
- latchTime auto adjusted based on configured lightIDs
- switchOffOnBlack working again
- compatible with led device component on/off
- more logging what's actually going on
- lightID detection should be better, i hope that all bridges supports this (update to recent firmware to be save)
- 2 new lamp models added

@ntim 
xy == CiColor::BLACK -> rgb 0,0,0 is never CiColor 0,0,0, i got 0.164, 0.04, 0 i switched now to check just against bri, that seems to work well.

Testers wanted
@Gamer06 @b1rdhous3 @seblagaffe

https://github.com/hyperion-project/hyperion.ng/issues/466
https://github.com/hyperion-project/hyperion.ng/issues/429

### DIFF
--- a/libsrc/leddevice/dev_net/LedDevicePhilipsHue.cpp
+++ b/libsrc/leddevice/dev_net/LedDevicePhilipsHue.cpp
@@ -4,10 +4,6 @@
 // qt includes
 #include <QtCore/qmath.h>
 #include <QNetworkReply>
-#include <iostream>
-
-const CiColor CiColor::BLACK =
-{ 0, 0, 0 };
 
 bool operator ==(CiColor p1, CiColor p2)
 {
--- a/libsrc/leddevice/dev_net/LedDevicePhilipsHue.cpp
+++ b/libsrc/leddevice/dev_net/LedDevicePhilipsHue.cpp
@@ -351,6 +351,7 @@ LedDevicePhilipsHue::LedDevicePhilipsHue(const QJsonObject& deviceConfig)
 	_deviceReady = init(deviceConfig);
 
 	connect(&bridge, &PhilipsHueBridge::newLights, this, &LedDevicePhilipsHue::newLights);
+	connect(this, &LedDevice::enableStateChanged, this, &LedDevicePhilipsHue::stateChanged);
 }
 
 LedDevicePhilipsHue::~LedDevicePhilipsHue()
@@ -449,14 +450,10 @@ int LedDevicePhilipsHue::write(const std::vector<ColorRgb> & ledValues)
 	return 0;
 }
 
-int LedDevicePhilipsHue::switchOff()
+void LedDevicePhilipsHue::stateChanged(bool newState)
 {
-	lights.clear();
-	return 0;
-}
-
-int LedDevicePhilipsHue::switchOn()
-{
-	bridge.bConnect();
-	return 0;
+	if(newState)
+		bridge.bConnect();
+	else
+		lights.clear();
 }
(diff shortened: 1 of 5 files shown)


brindosch:
I think there is a brightness calculation issues of blue (at least)
red (255,0,0) leads to brightness  72/254 (!)
green (0,255,0) leads to brightness 170/254
blue (0,0,255) lead to brightness 12/254 (!)

You can watch the brightness slider at the PhilipsHue App for better traceability. Don't forget to set the brightness compensation to 0% on the color page if you didn't used my webui philipshue assistant for configuration (available at led hardware -> philipshue)

So please report your results.

redPanther:
Great! I wished I had a hue ....

libsrc/leddevice/dev_net/LedDevicePhilipsHue.cpp
```diff
@@ -126,95 +122,134 @@ float CiColor::getDistanceBetweenTwoPoints(CiColor p1, CiColor p2)
...
+				Error(log, "Got invalid response from bridge");
+				return;
+			}
+			// check for authorization
+			if(doc.isArray())
+			{
+				Error(log, "Authorization failed, username invalid");
+				return;
+			}
+			else
```
redPanther: Code lile
If ()
{
 return
}
else
{
action
}

Can be written as:
If ()
{
Return
}

action

No need fot 'else' branch

brindosch: Of course, now where you name it

brindosch:
@redPanther 
I added a new LedDevice::switchOn() method, as i want a new initialization if you enable the led device. This removes a bool workaround in my write() function
Hope this is ok

redPanther:
Hmm I'm not sure this is the best solution. Perhaps it is better to define signals when enable state is changed. Then connect to the signals if needed

brindosch:
You mean one signal (and pass the bool state)?
The smartest way to get a pointer to the base class is?
 - static_cast at the child device constructor: LedDevice* md = static_cast<LedDevice*>(this);
 - method at base class that returns a pointer to it (eg getInstance())

Right?

redPanther:
no, I mean:
```
void LedDevice::setEnable(bool enable)
{
	// emit signal when state changed
	if (_enabled != enable)
	{
		emit enableStateChanged(enable);
	}

	// set black to leds when they should go off
	if ( _enabled && !enable)
	{
		switchOff();
	}
	_enabled = enable;
}
```

and add: `signal  enableStateChanged(bool);` in header.

in derived class you can connect to this signal with a slot. This is a bit cleaner and means more what you want. You can use this signal outside the leddevices too. E.g. when a leddevices deactivates itselfe due to an error (e.g. network unreachable ...)

brindosch:
So you want to keep the switchOff function? The signal replaces the functionality!? The write(BLACK) isn't a big deal.

To be more clear, to connect to the signal i need a pointer of the base class(?)
This works, but is it the way to go? esp as child class..
```
	LedDevice* ld = static_cast<LedDevice*>(this);
	connect(ld, &LedDevice::ledDeviceStateChanged, this, &LedDevicePhilipsHue::stateChanged);
```

redPanther:
`connect(this, SIGNAL(enableStateChanged(bool)), this,  SLOT(stateChanged));`
cast should not needed because we in a derivided class.

I want to keep switch off func. (or rename to setBlack ...).

brindosch:
Hm, now i ask myself why i didn't tried a second this :) But makes sense. Will do so... :)

brindosch:
Thank you for your feedback

PR #486, opened 2017-11-19
Move WebSocket to Webserver & HttpJsonRpc async

brindosch:
**1** Tell us something
- Both well tested, i don't expect issues

## Breaking Change!
WebSocket has now the same port as the webserver, it is no longer reachable at the json port


brindosch:
Pi debian8 passed.

Issue #490, opened 2017-12-17
Connection to KODI 18 broken


gismo2004:
According to  [xbmc github](https://github.com/xbmc/xbmc/pull/12281) and [forum](https://forum.kodi.tv/showthread.php?tid=324598) Kodi does not allow "WriteAccess" via "GET" methode anymore.

Thus using JSONP will not work, because it is using GET for all requests. [link:](https://stackoverflow.com/a/16416877)

unsing JSON leads to problems with ["Access-Control-Allow-Origin"](https://stackoverflow.com/a/46955558) when directly accessing the API. As far as i understand, using a proxy is the only way to handle this (as written in the last link). 

Unfortunately im not skilled enough to create such thing. 

if i can help somehow, please let me know.

cheers

brindosch:
According to the PR this is fixed?
Closed until someone says something different :)

Issue #474, opened 2017-09-15
[Testers wanted] Hyperion-Control for Kodi


brindosch:
Hello everyone, i am pleased to provide you the first testing version of a new Kodi Addon for Hyperion.
**This addon will replace the KodiCheck at Hyperion!**

Current functions:
- Enable or disable a Hyperion component of your choice based on the conditions you already know from KodiChecker
- Enable 2D/3D switch (has a little bug due to missing python apis / race condition with json rpc)
- Enable Hyperion on startup / disable on shutdown
- Settings reload during runtime
- Hyperion Server detection (SSDP)
- Kodi 17/18

Flaws:
- http! No detection of hyperion restarts, move to websocket or raw tcp socket.

Please install with "install from zip file"
Don't forget to leave some feedback
[script.service.hyperion-control-1.0.0.zip](https://github.com/hyperion-project/hyperion.ng/files/3412180/script.service.hyperion-control-1.0.0.zip)

redPanther:
perhaps use that:
https://pypi.python.org/pypi/websocket-client/

brindosch:
I already tried that to implement, but failed. I got no reponse/handshake, exception. No clue what i am doing wrong here. But this is my first python "app" so i have no experience which may result in simple issues. Maybe it's a lifetime issues. You can have a look at connect_ws.py which was a test.

Anyway, a huge issues are loops that freezes everything, esp the send/recv. They need to quit when kodi requests it, so that the script can properly exit / perform exit tasks.

Some things i forgot to mention
- i forgot pictures mode!
- kodi supports a reload of settings during runtime, this is full implemented, but kodi returns always old values, no clue why, need to ask kodi pros. (That worked earlier in development )

redPanther:
Will test it asap.
On our forum I had posted pycode for raw connection. I fon't know where exectly, but perhaps you find the code

brindosch:
Hello again,
updated first post with new zip (v0.0.2).
Changes
 - Code cleanup
 - startup wizard (sadly it's not that what i intended for, because of a issue with setSetting(). Maybe find the issue)
 - fixed read settings during runtime

@redPanther 
I didn't found your post, can you remember at least in which area you posted it or topic it was related to?
the whole socket stuff is not as easy as i thought if you want it as clean as possible with async send/read, non blocking. Like a QTcpSocket :)

brindosch:
### Update - V0.0.3
-Can be installed on Kodi 18

[...]

PR #492, opened 2018-01-06
Use QNetworkAccessManager to handle Kodi requests

gismo2004:
Kodi does not allow to send "commands" via method GET and using javascript leads to "Access-Control-Allow-Origin", thus create a proxy to handle our requests.

I'm no coder ;-)

-
https://github.com/hyperion-project/hyperion.ng/issues/490


brindosch:
Thank you for your PR. It seems you are a coder - now :)
Sad that kodi does no longer accept these request. While i created this it was more a bonus. Using Hyperion as forwarder is possible, but i dislike it. Another idea is to use Kodis Websocket capabilities which is implemented at all browsers (http://kodi.wiki/?title=JSON-RPC_API#WebSocket)

That way you can send json formatted messages to Kodi without any restrictions from the ui (http://kodi.wiki/view/JSON-RPC_API/v8)

May you probably have a look?

Thank you
brindosch

gismo2004:
That sounds like a new challenge :-D --> will definitely try to get this working. 

But it will take some time, because of family work and so on. So if it is no matter of time, i will push a new PR some day...

BTW: i'm from Austria, so if you prefere to write Deutsch... 
cheers

brindosch:
Great, time doesn't matter.
If you have questions or need hints just ask. We can switch to german, in case it's getting too technical :)

Thank you
brindosch

MartB:
I used the kodi websocket in the past it is pretty powerful im wondering if we should implement it into hyperion to get additional informations like playback start/stop events without the need for the kodi addon ?

brindosch:
Well, the situation is this

- kodicheck has been removed from Hyperion, because there is no reason to provide a _optional_ part at the core which can be simply controlled now via API
- The replacement for this is available now as Kodi Addon and will be pushed to the kodi repository during release https://github.com/hyperion-project/hyperion.ng/issues/474
- However, Kodi is not the only software that wants to be used together with Hyperion. It is currently impossible to install optional services that build a bridge between your wanted service and Hyperion or any other simple eventing like time based triggers. If think it's out of discussion to blow up the core for every task the user wishes. Which is the reason why i implemented a simple plugin system. You've probably seen it at my rework branch.

This pr wants to implement a picture pusher to be used together with the ui calibration wizard.

MartB:
We could technically merge the kodi img grabber and your addon from #474 in a later stage when the reworks on the server are done. However about that websocket part, i have an implementation lying around for this from an old project that i could adapt.

However the websocket api seems to not display the next image until the user manually presses the right mouse button (whatever that action is in kodi).

@gismo2004 did you encounter the same issue ?

gismo2004:
Hi,

sorry for the late response. I had some time to play around with WS. 
It works in principle.

But the current implementation is definitely not the final one. I need to find a way to create a kind of semaphore, since all that stuff is async and i don't want to block the scripts with the requests to Kodi.

Please let me know, what you think about the current stuff, and feel free to blame everything you don't like. as i said, I'm no coder, and i just learn how to work with stuff like that :-)

Cheers

MartB:
That does look like a solid start already but yeah we need to do this in a event driven way.

I adapted some old scripts of mine for hyperion ng they are not the best but maybe they inspire you to something new haha.

https://github.com/MartB/hyperion.ng/commit/4c7ccb756ebb0fe86314ce69882736e63300f5c4

gismo2004:
@MartB i would say i don't need a inspiration --> looks like pushing your commit makes much more sense and less work ;-)
Is there a reason why not using yours?

MartB:
@gismo2004 i didnt want to steal your pull request 🙈 
The connection logic is a bit wonky (error handling) and it could probably use some further testing.
If you see any room for improvement go ahead and comment on my code and when i have time i will get to it.

gismo2004:
@MartB My pull request has been created with the knowledge: Kodi is not working. Now that i see you already have a very nice implementation, and i am far away from creating cool stuff like this, you definitely not stealing something :-D 
Means, that i hope you are pushing your stuff, because it works as expected. There is just one thing i have seen, and i have commented on your commit.

If all agree with MartB's implementation i would like to close this PR?

cheers

RickDB:
@gismo2004 just to be sure we can close this one without merging?

gismo2004:
@RickDB YES sir :-)

PR #503, opened 2018-03-01
Apa104

penfold42:
Added support for apa104 LEDs. very similar to ws2812 but different timing

### DIFF
--- a/assets/webconfig/js/content_leds.js
+++ b/assets/webconfig/js/content_leds.js
@@ -470,7 +470,7 @@ $(document).ready(function() {
 	
 	// create led device selection
 	ledDevices = serverInfo.ledDevices.available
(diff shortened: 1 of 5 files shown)


penfold42:
This might not be necessary as the ws2812 driver is reported as working with the apa104.
https://hyperion-project.org/threads/apa-104-support.2729/

WoLpH:
As far as I can gather they are really similar (the APA104 specsheet is pretty much fully copied from the WS2812) and I don't think there is much point to this. Perhaps @RickDB can shine a light on this since he's got a few APA104s

RickDB:
Will double check it once my APA104s arrive (next week hopefully) as I currently using the older ones which is the 2-wire AP102s.

tpmodding:
penfold knows what he does...code should be fine :)

RickDB:
Alrighty then 👍

RickDB:
Always nice to have it in a different class even if the specs are similar to WS2812 so that if we need to tweak APA104 can always do so.

tpmodding:
absolutely right

PR #514, opened 2018-08-04
Update screen dimensions only if they really change

oCanna:
The screen and output image resolution don't need to be updated on every frame/grab.

### DIFF
--- a/libsrc/grabber/x11/X11Grabber.cpp
+++ b/libsrc/grabber/x11/X11Grabber.cpp
@@ -119,8 +119,7 @@ int X11Grabber::grabFrame(Image<ColorRgb> & image, bool forceUpdate)
 {
 	if (!_enabled) return 0;
 
-	if (forceUpdate)
-		updateScreenDimensions(forceUpdate);
+	updateScreenDimensions(forceUpdate);
 	
 	if (_XRenderAvailable && !_useXGetImage)
 	{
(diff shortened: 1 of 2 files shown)


RickDB:
If I read the change correctly this will always call updateScreenDimensions() on every frame no matter what as it removes the forceUpdate check, that specific function does run a few things even if not forced to update like log dimension and scaling checks.

What is use case where this is necessary to do on each frame event?

oCanna:
It doesn't remove the forceUpdate trigger entirely. It is still present in the second if statement inside the updateScreenDimensions() function. And yes, this function is called on every single frame, but that was also the case with the original code because the call of this function was only reliant on the state of the forceUpdate boolean which was always true.
Now, the function is called on every single frame, no matter what state this boolean has. But the whole screen dimension update mechanism is only executed if the screen dimensions have really changed or if the forceUpdate boolean is true, which now ist set to false by default. So with my PR, screen dimension updates are only performed on startup or if the dimensions change. Without this modification the whole updateScreenDimensions() function is executed if you look again at the second if statement inside this function because of the forceUpdate boolean was set to true by default. This is unnecassary and consumes only additional cpu time.

I hope it is clear now and sorry for my bad english.

RickDB:
No problem and explanation is nice and detailed 👍 
Gonna merge it and if you have anymore changes feel free to contribute.

PR #519, opened 2018-11-06
Fix - Philips Hue RGB Brightness Color Conversion

SJunkies:
Hue RGB Color Conversion for Brightness is wrong! - use RGB to HSV instead of XYZ Space!
For correct use, reset brightnessFactor back to 1.0 in config! - It was just a Workaround, not the Solution!


RickDB:
Let me know if you want me to merge or wait for further testing / additional patches, don't own any Hue products anymore which I can test on so if some other team member can test this that would be great.

RickDB:
Hi,

We had a recent rework merged in which broke this PR sadly, could you please rebase it on our current master?

PR #533, opened 2019-02-07
Fixes PhilipsHue Brightness, ColorGamut + new Lights

SJunkies:
Hue RGB Color Conversion for Brightness is wrong! - use RGB to HSV instead of XYZ Space! 
For correct use, reset brightnessFactor back to 1.0 in config! - It was just a Workaround, not the Solution!
Fixed wrong ColorGamut A/B/C Values + Added New Lamp/Light Models: LCT015, LCT016, LCT024


RickDB:
Thanks for the fix and recreating the PR 👍

SJunkies:
you're welcome! :)

PR #534, opened 2019-02-08
fix amlogic flatbuffer library

Portisch:
With this change the flatc compiler will be found on amlogic

### DIFF
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 set(USE_SYSTEM_FLATBUFFERS_LIBS ${DEFAULT_USE_SYSTEM_FLATBUFFERS_LIBS} CACHE BOOL "use flatbuffers library from system")
 
 if (USE_SYSTEM_FLATBUFFERS_LIBS)
-	find_package(flatbuffers REQUIRED)
+	find_program(FLATBUFFERS_FLATC_EXECUTABLE NAMES flatc REQUIRED)
 	include_directories(${FLATBUFFERS_INCLUDE_DIRS})
 else ()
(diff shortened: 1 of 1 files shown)


Portisch:
Also here is a patch to add libarys to fix compiling for amlogic. But I am sure this will be device specific and can't be include with a pull request, isn't it?
https://gist.github.com/Portisch/1926ffab6b543d72792cf32da9e1cbb9

RickDB:
Not sure if it's possible to make the CMAKE file selective for certain platforms as pretty sure it will break other platforms if we include just the extra-linking-libraries.patch.

Portisch:
I am also not sure. If it will not be accepted I add a local patch file.
I just tried the newest Hyperion version without any config change and it didn't work. So more testing is required.

RickDB:
Want me to merge this one already and optionally that patch in a separate PR later?

Portisch:
Without this patch it won't compile on Coreelec anyway.

RickDB:
Gonna ping @Paulchen-Panther then to check if there's way to include this patch.

RickDB:
What you could do in the meantime is include the patch in this PR, Travis will let's know if it compiles properly and can do local testing afterwards as an additional check.

Paulchen-Panther:
> Also here is a patch to add libarys to fix compiling for amlogic. But I am sure this will be device specific and can't be include with a pull request, isn't it?
> https://gist.github.com/Portisch/1926ffab6b543d72792cf32da9e1cbb9

Yes, it's possible
```
if (ENABLE_AMLOGIC)
target_link_libraries(${PROJECT_NAME} pcre16 dl z)
endif ()
```

Portisch:
I have done a new "compile fix" PR: https://github.com/hyperion-project/hyperion.ng/pull/536

PR #536, opened 2019-02-12
fix amlogic compile

Portisch:
added missing librarys and inludes by amlogic compiler flag


tpmodding:
all checks passed :-) looks good!

RickDB:
Code looks fine and travis agrees so merging 👍

PR #537, opened 2019-02-13
fix V4L2 grabber enable

Portisch:
re-enable the "enable" config flag for V4L2 grabber

V4L2 would start everytime, also if enabled == false

### DIFF
--- a/include/grabber/V4L2Wrapper.h
+++ b/include/grabber/V4L2Wrapper.h
@@ -24,6 +24,7 @@ public slots:
 	void setCropping(int cropLeft, int cropRight, int cropTop, int cropBottom);
 	void setSignalDetectionOffset(double verticalMin, double horizontalMin, double verticalMax, double horizontalMax);
 	void setSignalDetectionEnable(bool enable);
+	void setDeviceVideoStandard(QString device, VideoStandard videoStandard);
 
 private slots:
 	void newFrame(const Image<ColorRgb> & image);
--- a/libsrc/grabber/v4l2/V4L2Grabber.cpp
+++ b/libsrc/grabber/v4l2/V4L2Grabber.cpp
@@ -54,9 +54,6 @@ V4L2Grabber::V4L2Grabber(const QString & device
 {
 	setPixelDecimation(pixelDecimation);
 	getV4Ldevices();
-
-	// init
-	setDeviceVideoStandard(device, videoStandard);
 }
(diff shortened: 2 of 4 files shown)


b1rdhous3:
Hi, 

how can I test that the enable is broken?
When I try to disable and enable the v4l grabbing the logs looks good:
[hyperiond HYPERION] (DEBUG) (PriorityMuxer.cpp:153:registerInput()) Register new input 'System/V4L' with priority 240 as inactive
 [hyperiond ComponentRegister] (DEBUG) (ComponentRegister.cpp:67:componentStateChanged()) V4L capture device: enabled
 [hyperiond HYPERION] (DEBUG) (PriorityMuxer.cpp:330:setCurrentTime()) Set visible priority to 240
 [hyperiond WEBSOCKET] (DEBUG) (WebSocketClient.cpp:197:sendClose()) send close: 1000 
 [hyperiond WEBSOCKET] (DEBUG) (WebSocketClient.cpp:32:WebSocketClient()) New connection from ::ffff:192.168.178.21
 [hyperiond WEBSOCKET] (DEBUG) (JsonAPI.cpp:916:handleLoggingCommand()) log streaming activated for client ::ffff:192.168.178.21
 [hyperiond WEBSOCKET] (DEBUG) (JsonAPI.cpp:925:handleLoggingCommand()) log streaming deactivated for client  ::ffff:192.168.178.21
 [hyperiond HYPERION] (DEBUG) (PriorityMuxer.cpp:248:clearInput()) Removed source priority 240
 [hyperiond HYPERION] (DEBUG) (PriorityMuxer.cpp:330:setCurrentTime()) Set visible priority to 255
 [hyperiond ComponentRegister] (DEBUG) (ComponentRegister.cpp:67:componentStateChanged()) V4L capture device: disabled

Also if I check the config before stopping hyperion  "v4lEnable": false is set and after restarting hyperion this property is still set to false.

Portisch:
Right, this was the reason I found this bug. V4L2 was enabled for amlogic. But it didn't break hyperiond because in the config the default for enable is false. So V4L2 got loaded, but not started.
In the newer versions V4L2 get startet even if enable=false and this will crash on amlogic devices when included in the compile.
In the new CE package V4L2 will not be included anymore.

b1rdhous3:
Ah okay, understood. Seems reasonable to me.

PR #552, opened 2019-04-13
Support additional Nanoleaf devices

Lord-Grey:
- Refactored Aurora device support to cover multiple Nanoleaf devices
- Added Nanoleaf device discovery
- Updated SSDPDiscover to be generic for given services
- Support devices which allow to be switched on/off physically, rather than have a black output stream.
Devices were not switched on after disabling Hyperion or switching LEDs off before.

Instructions to create Authentication Token key is in http://forum.nanoleaf.me/docs/openapi under Authorization using REST or https://github.com/software-2/nanoleaf.

// NANALEAF DEVICE CONFIGURATION 
"device" : {
    "type"   : "nanoleaf"
    "output" : "_hostname or IP_",      // Optional. If empty, device is tried to be discovered
    "token"  : "_Authentication Token_",
},


RickDB:
Looks good 👍  , ready for merge?

Issue #542, opened 2019-02-25
Impossible to use hyperion-v4l2


washcroft:
Following @Paulchen-Panther 's latest updates, it is no longer possible to use the hyperion-v4l2 utility as the device name is never set:

```
# hyperion-v4l2 --device /dev/video0

hyperion-v4l2:
        Version   : 2.0.0 (master (GitHub-c2d19063/91054d8d-1550176449) Git Remote: https://github.com/hyperion-project/hyperion.ng.git)
        build time: Feb 24 2019 16:59:47
[hyperion-v4l2 V4L2:/dev/video0] <ERROR> Throws error nr: Cannot identify '' error code 2, No such file or directory
[hyperion-v4l2 V4L2:/dev/video0] <ERROR> Throws error nr: VIDIOC_QUERYCAP error code 9, Bad file descriptor
[hyperion-v4l2 V4L2:/dev/video0] <ERROR> Throws error nr: close error code 9, Bad file descriptor
```

Caused by this commit where the device/input number (from the constructor parameters) are now ignored:

https://github.com/hyperion-project/hyperion.ng/commit/2a77f6f0122691fe1961259178c1ce9d0caca053#diff-aec2cfe446c748ff341302e49e490d6bL33

The V4L2 grabber only works within Hyperion because of the call to setDeviceVideoStandard() which also sets _deviceName.

Paulchen-Panther:
I hope it has been fixed in PR #539 . Thanks for your bug report

vansoest:
I’ve tried your BugFixes2 branch but it hasn’t fixed the issue:

```
# hyperion-v4l2 -d /dev/video/fushicai_usbtv007
hyperion-v4l2:
	Version   : 2.0.0 (makepkg (GitHub-e4bea694/69ceee8b-1554220455) Git Remote: /home/mediacenter/.cache/yay/hyperion.ng-git/hyperion.ng)
	build time: Apr  2 2019 16:35:30
[hyperion-v4l2 V4L2:/dev/video/fushicai_usbtv007] <ERROR> Throws error nr: VIDIOC_S_INPUT error code 22, Invalid argument
[hyperion-v4l2 V4L2:/dev/video/fushicai_usbtv007] <ERROR> Throws error nr: VIDIOC_STREAMON error code 16, Device or resource busy
[...]

PR #567, opened 2019-06-05
Bug fixes, new implementation ...

Paulchen-Panther:
_Changes_:

- V4L2 video format "MJPEG" implemented (libjpeg/qimage)
- Inactive priorities are now correctly skipped by the Priority Muxer (https://hyperion-project.org/threads/falsches-bild-wird-weitergeleitet.3594)
- If the internal v4l2 grabber is switched off, it will now be de-initialized (Issue: #542)
- Extended error detection in MJPEG decompress method
- Default instance capture activation if it is missing in the configuration (https://hyperion-project.org/goto/post?id=13656#post-13656)
- Fixed some small (think)mistakes (https://hyperion-project.org/threads/hdmi-grabber.3302/page-2#post-14887)
- Code cleanup. Mostly line endings and indentions
- The default Boblight priority has been set to 128. If the sent priority differs from the default priority, it is registered (https://hyperion-project.org/threads/hyperion-ng-enigmalight-boblight-server.3558/#post-15006)
- Submodule flatbuffer updated to current master
- Travis linux OS upgraded to Xenial
- Increases linker performance under Linux builds by using Gold linker, if available
- CCache is used if available
- Removed statistic class (Stats.cpp) from project due to the missing result (sorry @Brindosch)
- Added LGTM bandges to README for code analysis overview
- Added Azure Pipeline build check.

~~Unfortunately I could not fix the timeout problem for the Travis job "AMLogic" yet. I hope you all forgive me.~~
The actual problem with Travis I have now solved by adding the Azure CI service. Azure allows a longer build time (unlimited) than Travis. Now the question is what to do with Travis!?!

@RickDB , @WoLpH , @tpmodding , @penfold42 , @b1rdhous3 
Any suggestion?

[Build Status](https://dev.azure.com/Hyperion-Project/Hyperion.NG/_build/latest?definitionId=2&branchName=master)


RickDB:
Think we should move off Travis for now at least at least as it's messing up our validation for PRs, noticed the Azure CI API requests and will approve them so we can give Azure a test run first.

WoLpH:
It seems to be mostly working. The only issue right now is the ARM8 build timing out because it's too slow. I'm trying a slightly different concurrency setting now which might help. If not, we can always ignore or even disable that build.

Disabling Travis entirely for 1 broken build in the entire job seems excessive :)

WoLpH:
Upon further inspection the main problem seems to be the performance. It looks like travis is working just fine except for being at the edge of timing out.

Perhaps a `MAKEFLAGS='-j 4'` or so will already fix the issue. I'll have to run some tests. But ccache might also fix the issue.

Paulchen-Panther:
@RickDB 

> Noticed that already went well so let's migrate +1 , what steps do we need to follow for this?

The link to the badges in the README.md then had to be adjusted. Azure would switch to the main repository and it's already running.

@WoLpH 
Personally, I have nothing against Travis but if I look at the times Travis needs for armv6 / armv7 Cross Compiling, then it may be (as the project grows) that these tests will fail as well.
We could start by deactivating the armv8 architecture.

> Perhaps a MAKEFLAGS='-j 4' or so will already fix the issue. I'll have to run some tests. But ccache might also fix the issue.

I believe the tests are already running with the makeflag -j 2. [Link](https://github.com/hyperion-project/hyperion.ng/blob/master/.travis/travis_build.sh#L20)

WoLpH:
You're completely right. We already have both the makeflags and ccache enabled so I'm not sure if much can be done on that side.

My main reason for liking Travis is that it has a bunch of easy integrations available that make for easy deployment. But the building (and debugging of the builds) is often pretty painful. And at the very best it's still slow.

For the ARM builds we should just drop Travis completely, it's completely pointless when it times out all of the time. We can still keep it for i386 and AMD64 but that's it.

RickDB:
It has been broken for a while though not just this PR, if we can fix it with a few multi-core flags would be great.

Paulchen-Panther:
I've disabled the ARM tests for now. If there are changes to the Travis speed you can always activate it again in the travis file.

RickDB:
Awesome, want to merge it right away?

Paulchen-Panther:
With pleasure :grinning:

WoLpH:
Excellente work! Thank you :)

Paulchen-Panther:
@RickDB 
Could you please check the WebHooks?

azure

RickDB:
Did enable Azure via the invite today but will double check to see if I missed anything.

Paulchen-Panther:
Thank you :+1:

RickDB:
Shows up as approved for Hyperion projects (ng / vanilla / docker):

Azure

Paulchen-Panther:
Is there a choice of webhooks?
azure

RickDB:
No webhooks listed there for both Hyperion.ng and the top level (Hyperion-project).

RickDB:
Correction see some but think those are by the retired admin:

Capture

Paulchen-Panther:
It seems like I do not have the necessary permissions to add this repository. Either you sign up in Azure itself and I send you an invitation link or you transfer the necessary rights.

RickDB:
Reset it just now, could you resend the invite?

Also upgraded your rights so have the same as @WoLpH at the moment.

Paulchen-Panther:
@RickDB Thank you.
It was apparently due to the privileges. I activated the rights on the third-party application access policy page and it worked right away.  :+1:

RickDB:
Nice 👍

PR #575, opened 2019-06-22
Fix systemd unit

hissingshark:
Variables like `%i` don't appear to be supported in systemd unit files.  This means a config file cannot be created and `hyperiond` fails to start.  It might be possible that some distributions have newer versions of systemd that have fixed this, but not all.

### DIFF
--- a/bin/service/hyperion.systemd
+++ b/bin/service/hyperion.systemd
@@ -1,11 +1,11 @@
 [Unit]
-Description=Hyperion ambient light systemd service  for user %i
+Description=Hyperion ambient light systemd service for user root
 After=network.target
 
 [Service]
 ExecStart=/usr/bin/hyperiond
 WorkingDirectory=/usr/share/hyperion/bin
-User=%i
+User=root
 TimeoutStopSec=5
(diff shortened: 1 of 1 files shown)


Paulchen-Panther:
@brindosch What do you say to this problem?

brindosch:
That's a placeholder which should be replaced by systemd. The idea is to select the current user as owner.
Did this task fail? And on which OS? It wasn't a problem with Ubuntu 16.04 and Debian 9.

hissingshark:
I'm running OSMC (Debian Stretch based).
`systemd --version`
>  systemd 232

I understood that it should be substituted by the user name, but that wasn't working.
It sounds like it was [reported and fixed](https://github.com/systemd/systemd/issues/679) in systemd a long time ago.  As I said, I assumed our version of systemd was too old for this method?

But wouldn't you know it, I've just tried to replicate it again to post the errors and it's ok.  We've just had a system update so perhaps that was part of it.  Still seems like it should have been a long time ago...

Thank you for your attention, but I'll close this now.

PR #576, opened 2019-07-02
Bugfix/Hotfix/Update

Paulchen-Panther:
- Bugfix:
	- Corrected the led view in the configuration editor
	- The message forwarder only forwards the active priority if system and V4L grabbers are active
	- The Fadecandy device is now working again

- Hotfix:
	- The repeated shrinking of the led vectors was taken from the main processing loop Hyperion :: update () to avoid crashes when specifying Hardware Leds (hardwareLedCount) in the configuration.

- Update:
	- To minimize conflicts in the main processing loop, mutex were introduced
	- The activation state of the black bar detection is only applied in case of priority change
	- The LED class (and their child classes) now receives the priority change signals from the priority muxer


Paulchen-Panther:
I'll merge this, is this ok?!?!

b1rdhous3:
Yep. From my end okay.

PR #580, opened 2019-07-13
Allow to switch on LED device, if currently disabled

Lord-Grey:
Reimplement to be able to switch on/off devices via the LED Device UI button.
Before the change devices will only be switched off, but not on again.

This is mainly relevant for devices which override the LedDevice switchOn/Off methods to do API calls
rather than have a black output stream.

### DIFF
--- a/bin/create_all_releases.sh
+++ b/bin/create_all_releases.sh
@@ -10,12 +10,12 @@ make_release()
 	PLATFORM=$2
 	shift 2
 
-	rm -r hyperion.ng-${RELEASE}
-	mkdir -p hyperion.ng-${RELEASE}
+	rm -r build-${RELEASE}
+	mkdir -p build-${RELEASE}
 	rm -r deploy/${RELEASE}
 	mkdir -p deploy/${RELEASE}
 
-	cd  hyperion.ng-${RELEASE}
+	cd  build-${RELEASE}
 	cmake -DCMAKE_INSTALL_PREFIX=/usr -DPLATFORM=${PLATFORM} $@ -DCMAKE_BUILD_TYPE=Release -Wno-dev .. || exit 1
 	make -j $(nproc)  || exit 1
 	#strip bin/*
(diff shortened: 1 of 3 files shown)


b1rdhous3:
On my side the LEDs are switching off and on again without any code changes. 
Which grabber do you use? I used OSX Platform Capture for this testing. 

Anyways: I checked out your changes and compiled without errors. Also I can switch off and on leds. So on my side there are no real changes.

Am I am missing something?

Lord-Grey:
@b1rdhous3: The problem only occurs, if one overrides the switchOn/switchOff methods for a specific LED device where on/off is implemented via API-Calls (e.g. Nanoleaf).
The device will current switch off, but due to the fact that a state change will not trigger switchOn, it stays off.
The change is just the symetric implementation of switching on/off methods.

b1rdhous3:
Ah I see. Symetric implementation of this is making sense. 
I would approve this. (If this matter, haha)

Paulchen-Panther:
> The device will current switch off, but due to the fact that a state change will not trigger switchOn, it stays off.

Please use the public signal enableStateChanged (bool new state) to turn the device on or off. The reason is the limitation of the LED device threads.
Information Links:
https://stackoverflow.com/questions/20051224/qt-socket-notifiers-cannot-be-enabled-from-another-thread

Lord-Grey:
a) > limitation of the LED device threads.
What are the limitations of the LED device thread?
I am currently running the switch on/off API calls via the LedDeviceNanoleaf object's QNetworkAccessManager and it works as expected.
I am new to QT5. When would it fail?

b) Looking at the definition & description in LedDevice.h, it is not intuitive why one should use direct method calls in some scenarios and emit signals in others.

```
public:

	/// Switch the leds off (led hardware disable)
	virtual int switchOff();
	/// Switch the leds on (led hardware enable), used if reinitialization is required for the device implementation
	virtual int switchOn();
```
As per your description, shouldn't switchOn/Off emit the signal and overriding be not allowed to ensure one way of dealing with the scenario?

Paulchen-Panther:
@brindosch Was sagst du zu der Thematik? Wäre es nicht besser die virtuellen Funktionen in virtuelle slots umzuformen?

brindosch:
Yes/No, i think i wanted to merge on/off into one slot with bool arg instead of two methods. But i think not all led devices where ready for that step. This is just what i remember (2/2018)

I think there is no real reason to use 2 methods or slots?

Paulchen-Panther:
Ready to merge?

Lord-Grey:
Good to go!

PR #589, opened 2019-08-07
Debian post-inst script fails to correctly create system service when multiple users are present

### LINKED ISSUE #588: Debian post-inst script fails to correctly create system service when multiple users are present
Latest debian build fails to correctly create the system service when multiple users are present in the result of `who`, this causes the service to reference an incorrect user alias and therefore the service is unable to start.

Issue seen on the latest build, using a fresh install of Raspbian Buster.

bennealon:
Fixes an issue where multiple users returned from `who` causes grep to be greedy and return a regex matched string containing all users, not the first user as intended.
This caused the package installation on Raspbian to create an incorrect service name which could not be started as there was a mismatch in user vs expected user.

As discussed with @Paulchen-Panther [HERE](https://hyperion-project.org/threads/installing-hyperion-ng.3491/page-3#post-15695)

Fresh install of Raspbian Stretch, calling `who` from the terminal returns two users. Doing the same on an SSH session returns 3:
```
pi@raspberrypi:~ $ who
pi       tty1         2019-08-07 17:38
pi       tty7         2019-08-07 17:39 (:0)
pi       pts/0        2019-08-07 17:41 (10.26.25.176)
```
After regexp:
```
pi@raspberrypi:~ $ who | grep -o '^\w*\b'
pi
pi
pi
```

Fix changes to:
```
pi@raspberrypi:~ $ who | grep -o -m1 '^\w*\b'
pi
```
Tested installing .deb on Raspbian Buster - now works. can someone please check the .rpm please, as i have made the same changes there.
Fixes #588

### DIFF
--- a/cmake/debian/postinst
+++ b/cmake/debian/postinst
@@ -29,7 +29,7 @@ BOOT_BERRYBOOT=$(grep -m1 -c '\(/var/media\|/media/pi\)/berryboot' /etc/mtab)
 NET_IP=`hostname -I | cut -d " " -f1`
 
 # search for users in system, returns first entry
-FOUND_USR=`who | grep -o '^\w*\b'` || "root"
+FOUND_USR=`who | grep -o -m1 '^\w*\b'` || "root"
 
 # determine if we should use a service
 ENABLE_SERVICE=0
--- a/cmake/debian/postinst
+++ b/cmake/debian/postinst
@@ -29,7 +29,7 @@ BOOT_BERRYBOOT=$(grep -m1 -c '\(/var/media\|/media/pi\)/berryboot' /etc/mtab)
 NET_IP=`hostname -I | cut -d " " -f1`
 
 # search for users in system, returns first entry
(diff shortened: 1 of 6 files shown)


bennealon:
I've been wondering whether we should be using `whoami` instead of `who`, as there can still be ambiguity over which user you want when there is multiple ones with `who`, which would not be the case with `whoami`.
Does anyone object if i change this, as i feel it is more correct in a multi-user system?

Paulchen-Panther:
> Does anyone object if i change this, as i feel it is more correct in a multi-user system?

I totally agree with your opinion.

bennealon:
> > Does anyone object if i change this, as i feel it is more correct in a multi-user system?
> 
> I totally agree with your opinion.

Great cheers, i'll change the PR 👍

bennealon:
`whoami` won't work, since you have to `sudo dpkg...` to install, so the user comes back as `root` 😞 , so i've reverted that change back to using `who`.

Paulchen-Panther:
Ready to merge?

bennealon:
Yep, go for it!

PR #601, opened 2019-08-14
Refactored Exceptions in main & PythonInit

Lord-Grey:
- Refactored exception handling in main
Exceptions thrown by HyperionDaemon did not result in stopping the application, but had the qt application still running. Exception thrown for rootPaths were even not caught, as thrown outside a try block. In case hyperion should run without daemon ,come back to me.

- During my testing to ship libraries with the app, I identified that hyperiond starts even, if Python failed to initialise. Added an exception, if Python failed initialisation.


brindosch:
Thank you :)

PR #610, opened 2019-08-18
Make origin in schema-image optional

ChaosCoder:
### 1. Tell us something about your changes.

The origin parameter when setting an image was set to required - unlike when setting a color or an effect. 

This change makes the origin parameter optional and sets the min and max length of the value to be the same as in color and effect.

### 2. If this changes affect the .conf file. Please provide the changed section

no

### 3. Reference an issue (optional)

none


Paulchen-Panther:
Excellent :+1:

brindosch:
Thank you

PR #623, opened 2019-09-08
Add PWM channel option for ws281x device in web interface

tripplet:
The ws281x device has a PWM channel [option](https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/leddevice/dev_rpi_pwm/LedDeviceWS281x.cpp#L33) which was missing from the web interface.


brindosch:
Thank you
@tripplet

PR #645, opened 2019-11-17
Embed QSQLITE driver at Amlogic platform

Portisch:
Embed QSQLITE driver on Amlogic platform into hyperiond binary by `Q_IMPORT_PLUGIN`

none

QSQLITE is not part of the system.
I am not sure if this would be needed on other platforms too.

### DIFF
--- a/libsrc/db/DBManager.cpp
+++ b/libsrc/db/DBManager.cpp
@@ -1,3 +1,6 @@
+// Hyperion includes
+#include <HyperionConfig.h>
+
 #include <db/DBManager.h>
 
 #include <QSqlDatabase>
@@ -7,6 +10,11 @@
 #include <QThreadStorage>
 #include <QUuid>
 #include <QDir>
+
+#ifdef ENABLE_AMLOGIC
+#include <QtPlugin>
+Q_IMPORT_PLUGIN(QSQLiteDriverPlugin)
+#endif
 
 // not in header because of linking
 static QString _rootPath;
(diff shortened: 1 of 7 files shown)


libsrc/db/DBManager.cpp
```diff
@@ -8,6 +11,11 @@
 #include <QUuid>
 #include <QDir>
 
+#ifdef ENABLE_AMLOGIC
```
Paulchen-Panther: Please change to #if ENABLE_AMLOGIC.
The variable ENABLE_AMLOGIC exists in any case and should only be checked for its value.

Portisch: Looks like this is not working without a compare.
So how to handle this? There are some more #ifdef like I used:
https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/grabber/framebuffer/FramebufferFrameGrabber.cpp#L49

Portisch:
I just checked it again. It have to be `#ifdef ENABLE_AMLOGIC`, When `ENABLE_AMLOGIC = 1` is set it's defined. If `ENABLE_AMLOGIC = 0` `ENABLE_AMLOGIC` is undefined.

Portisch:
Why is for `Build Debian Stretch (AMD64) packages` `ENABLE_AMLOGIC = ON`?
Same for `Build Debian Stretch (i386) packages`
qt-everywhere have to be build with `-qt-sqlite` to add the QSQLITE plugin

Portisch:
I think best is to include this as patch in our local build - closed!

PR #650, opened 2019-12-02
Add RS232 timeout for asynchronous connection

Portisch:
Add a timeout on asynchronous write. Sometimes it happen that the `.write` is successful but no bytes are written at all. This will be  catched by the timeout timer like the QSerialPort example is showing.

Also the handling of written bytes in function `bytesWritten` got modified as only one single transfer can be handled in one shot on the asynchronous connection.

none

See 1.


Paulchen-Panther:
Vielen Dank für deine Arbeit. Ich hoffe du bleibst uns erhalten. 👍

PR #654, opened 2019-12-14
Disentangle LedDevice/LinearColorSmoothing, Bug Fixes & Test support

Lord-Grey:

While decoupling ColorSmoorthing from LedDevice, I am coming across various bugs/challenges, as well as added additional capabilities to support testing during development. 

a) Fix MultiColorAdjustment coredump (see comment for TestScenario)
b) LedFileDevice: Add time delta between two writes (to validate update/write intervals)
c) LedDevice: Add method to write Led values to stdout
d) Add "Led Test" effect to have well defined switches between RGBWB to validate intervals and smoothing
e) Disentangle LedDevice and LinearColorSmoothing
- LedDevices now refresh themselves independently from LinearColorSmoothing
- LedDevices are now disabled by default and only after successful initialisation enabled
- LedDevices do not update(write), if not ready
- LedDevices write "Black", if switched-off (unless switch-off by LED hardware is supported)
- LedDevices, clear separation between threads of LedDeviceWrapper and LedDevices using signals and considering QT's thread affinity rules.
- LedDevices, new close() slot with BlockingQueuedConnection to ensure proper shutdown of devices before hyperion ends
- LedDevices, removal of throwing exceptions which are not caught properly between QT-Threads.
New error() method allowing to set a device in error and not being ready for usage.

To have proper thread alignment sequence per device should be as per original documentation

```
a. constructor  - without calling init()
b. start()
b.1 open()
b.1.1 init()
b.1.1.1 LedDevice::init()
b.2 setEnable(true), if init() was successful
....
c close()
c.1 LedDevice::close()
```
- Created LedDevice Code Templates to ease alignment to standard

- All LedDevices besides Phillips Hue were updated and aligned.
The updates include some fixing of cppcheck and clang findings

- Fixed, Effect configuration updates (effect configurations were appended and not updated)
- Fixed, Hyperion - Removed visiblePriorityChanged->update connection, as it results in double write()s
- Update, changed Framebuffer timer to "precise" rather than using a "coarse" timer per default

**Open topics/issues:** LinearColorSmoothing as not been fully reviewed and updated yet.
Disentangling of ColorSmoothing and Effects is required. Testing effects does not make use of the correct smoothing configuration, but the default one (0 record).


Lord-Grey:
Not sure why all already merged commits still show up in the list...
Everything after the last merge is new.

Lord-Grey:
In case I come across further issues/bugs, do you prefer to have individual issues capture, one collection issue created or no issues at all?
Not sure what is best practice here? Please advise.

Paulchen-Panther:
Da fehlt bestimmt ein `#include <chrono>` ???

Lord-Grey:
> Da fehlt bestimmt ein `#include <chrono>` ???
Interessant, lokal wird ohne Fehler kompiliert.
Nichtsdestotrotz, Danke für den Hinweis!

Paulchen-Panther:
@Lord-Gray

Fehlt noch was bei deinem PR? Kann ich mergen oder helfen?

Lord-Grey:
@Paulchen-Panther Ich bin gerade dabei die anderen Devices umzustellen, damit sie sich auch ins neue Schema einfügen und stabiler werden.

Paulchen-Panther:
Alles klar. Sag bescheid wenn du soweit bist. :+1:

Lord-Grey:
@Paulchen-Panther OK. Nach einem Refactoring Marathon von 66 Files können wir starten.

Nur die Phillip-Hue Files habe ich aktuell noch ausgelassen.

@SJunkies Hattest Du die Hue Files überarbeitet/vereinfacht? ggf. kannst Du sie dem Template nach anpassen [LedDevice Code template](https://github.com/Lord-Grey/hyperion.ng/blob/master/libsrc/leddevice/LedDeviceTemplate.cpp) oder mir zur Verfügung stellen.

Lord-Grey:
@All I had to update all LedDevice code files aligning them to QT's thread affinity rules.

May I ask you a favour?
Would you test the updated code against your LedDevices, please?
You may use my repository or the master, when Paulchen-Panther merged the PR.

Could you after testing comment on the test results providing the information per device tested?

Device Type:
Test (OK/NOK):
Smoothing (On/Off):
Rewrite (on/Off):

That may also help me going forward, knowing whom I can approach for testing support going forward...
Thanks for your support!

libsrc/hyperion/Hyperion.cpp
```diff
@@ -87,8 +87,9 @@ void Hyperion::start()
 		_ledStringColorOrder.push_back(led.colorOrder);
 	}
 
+	// NOTE: Removed visiblePriorityChanged->update connection, as it results in a double write()
 	// connect Hyperion::update with Muxer visible priority changes as muxer updates independent
-	connect(&_muxer, &PriorityMuxer::visiblePriorityChanged, this, &Hyperion::update);
```
brindosch: As far as i know this should catch the following:
- When a priority timeout ends (consider static color)
- or user requests a priority switch the leds won't update

The old solution was a timer for Hyperion:update without further conditions

Lord-Grey: @brindosch and others

OK. Let's figure out what happens with or without connect.
You may need to help me with additional test-cases...., esp. for 
> or user requests a priority switch the leds won't update

What are the steps to execute that testcase?

### Test

All tests with 
- a "file" device with timestamps on

- Smoothing off (to see the pure update() calls)

**With Connect (PriorityMuxer::visiblePriorityChanged -> Hyperion::update)**

a) Start of hyperion

tail: output.test: file truncated
2020-02-09 16:02:50.554 | +0024 [{67,253,33}{255,196,0}{193,0,228}{0,199,241}]
**2020-02-09 16:02:50.555 | +0000 [{67,253,33}{255,196,0}{193,0,228}{0,199,241}]**
2020-02-09 16:02:51.555 | +1000 [{70,253,31}{253,192,1}{190,0,229}{0,202,240}]

-> Duplicate record at the same time (+0000 is the delta since last Led write)

b) When a priority timeout ends (consider static color)
Set an Effect-Color (Red = rgb(255,0,0))

...
2020-02-09 16:09:33.399 | +0999 [{191,188,188}{207,207,207}{207,202,203}{179,166,173}]
2020-02-09 16:09:33.424 | +0024 [{255,0,0}{255,0,0}{255,0,0}{255,0,0}]
**2020-02-09 16:09:33.424 | +0000 [{255,0,0}{255,0,0}{255,0,0}{255,0,0}]**
2020-02-09 16:09:43.647 | +10223 [{189,185,185}{207,207,207}{128,110,122}{86,63,79}]

-> Duplicate record at the same time (+0000 is the delta since last Led write)

**Without Connect (PriorityMuxer::visiblePriorityChanged -> Hyperion::update)**

a) Start of hyperion

tail: output.test: file truncated
2020-02-09 16:12:15.143 | +0037 [{67,253,33}{255,196,0}{193,0,228}{0,199,241}]
2020-02-09 16:12:16.143 | +1000 [{70,253,31}{253,192,1}{190,0,229}{0,202,240}]

-> No Duplicate records

b) When a priority timeout ends (consider static color)
Set an Effect-Color (Red = rgb(255,0,0))

...
2020-02-09 16:13:31.899 | +0999 [{1&PriorityMuxer::visiblePriorityChanged, _ledDevice, &LedDevice::visiblePriorityChanged,82,178,179}{207,207,207}{214,215,216}{186,178,185}]
2020-02-09 16:13:32.722 | +0823 [{255,0,0}{255,0,0}{255,0,0}{255,0,0}]
2020-02-09 16:13:43.899 | +11176 [{183,179,180}{207,207,207}{128,110,122}{85,60,77}]

-> No Duplicate records

### Additional questions/findings

1. Are there other cases/scenarios that should be tested?
2. There is a connect in the DeviceWrapper (PriorityMuxer::visiblePriorityChanged ->&LedDevice::visiblePriorityChanged) which seems not relevant.
I suggest to remove that too, removing the tight coupling between components.

Lord-Grey: OK. Seems I found the scenario where no output is generated, if the connect is missing.

a) Set a fixed colour via the Systray as a new source.
-> Colour is written to the output
b) Switch to Platform Capture source
-> Output is updated
c) Switch to Fixed Colour source
-> No update is written
 
The duplicate writes in the other scenarios are not necessarily bad, but maybe there is a better alternative avoiding them. Let me have a look.

brindosch: We need lot's of tests that can run at ci.

A test should be:
- Set color at prio 1, set another color at prio 2, select prio 1 [Hyperion::setCurrentSourcePriority](https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/hyperion/Hyperion.cpp#L286)
No one calls Hyperion::update()
It should be possible to call update() on success like [Hyperion::setSourceAutoSelectEnabled](https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/hyperion/Hyperion.cpp#L280)
- Same with [Hyperion::clear()](https://github.com/hyperion-project/hyperion.ng/blob/master/libsrc/hyperion/Hyperion.cpp#L394)
- Last case: Timeout, a priority is dropped and therefore another priority is now active, who updates?

Lord-Grey:
For now, I  reverted the change "Remove Connect (PriorityMuxer::visiblePriorityChanged -> Hyperion::update) due to double writes"  to keep existing behaviour and not limit merging the PR (see [commit](https://github.com/hyperion-project/hyperion.ng/pull/654/commits/3cb27a01bfb7dbc30d721e6ba9c0da3446af6ef4))

libsrc/hyperion/Hyperion.cpp
Lord-Grey: @brindosch Would you mind outlining how I would do 

> Set color at prio 1, set another color at prio 2, select prio 1

via the UI? I only have the capture source and everything else gets prio 1 assigned.

Or are you describing just automatic test-cases? 

Additional questions:

When I manually set an Effect-Color or Picture Streaming via the Browser, both get always prio 1 and refresh with 10hz. Is it possible to change the refresh rate or is it always defaulted?
In this context, the UI states "Duration 2s", but the streaming does not end after 2s. What does the "2s" statement mean?

Lord-Grey:
@Paulchen-Panther Hi, wie sieht es aus, willst Du mal mergen?

Ich bin erstmal fertig und will auch den PR nicht noch größer werden lassen, damit Dir die Merges nicht noch komplexer werden. Wenn ich den neuen PR von brindosch sehe, kommt da ja noch ein "dicker Bocken".
 Fixes können wir ja dann über kleinere PRs integrieren.

libsrc/hyperion/Hyperion.cpp
brindosch: >> Set color at prio 1, set another color at prio 2, select prio 1

> via the UI? I only have the capture source and everything else gets prio 1 assigned.

Not possible, as the prio can't be configured at the ui

> When I manually set an Effect-Color or Picture Streaming via the Browser, both get always prio 1 and refresh with 10hz. Is it possible to change the refresh rate or is it always defaulted?

Is defaulted, the screen capture consumes lot's of memory. I still need to figure out the bottleneck here.

> In this context, the UI states "Duration 2s", but the streaming does not end after 2s. What does the "2s" statement mean?

It reflects the timeout from PriorityMuxer (which is set for each single image). This is visual hidden für other implementations (proto/flat). We may need to distinguish between an "image" and an "imagestream". Especially as proto/flat is just a protocol and nothing "special". A cleanup of the components/PriorityMuxer to outline the belongings better is required. But it's not always easy. An effect for example can be a "ledColorsStream" but also an "imagestream"

Lord-Grey:
@brindosch Thank you! Fingers crossed that not too many issues are popping up....

libsrc/hyperion/Hyperion.cpp
Lord-Grey: > It reflects the timeout from PriorityMuxer (which is set for each single image). 

The current "Duration 2s" on the UI created a different expectation to myself as a user rather than a timeout. Maybe the text is better removed from the screen to avoid confusion?

brindosch: Yes, the problem is that you can set a single image with timeout, and this timeout will be lost, as both are going through the same api method and there is no flag to mark it accordingly.

brindosch:
Regarding test i mean writing test for all classes where we expect a specific result. We are often caught in regressions due to things you can't see. Often a test "framework" is involved for simple code writing.

This can be for example the API class: "Test all methods to provide an access denied when not authorized".

Lord-Grey:
> Regarding test i mean writing test for all classes where we expect a specific result. We are often caught in regressions due to things you can't see. Often a test "framework" is involved for simple code writing.

That is understood and the way how are things done today.I only wanted to confirm that my understanding was right. We are on the same page.

libsrc/hyperion/Hyperion.cpp
Lord-Grey: > Yes, the problem is that you can set a single image with timeout, and this timeout will be lost, as both are going through the same api method and there is no flag to mark it accordingly.

Maybe I did not express myself clearly...

As a hyperion user, I do not find the duration information rather confusing than helpful:
https://snipboard.io/du20Jw.jpg

It gives too much room for interpretation what the time value may be.
I was just looking from a usability perspective, not with my developer lenses on. :)

Issue #674, opened 2020-02-12
Backlight threshold is not correct


T-IOOO:
I have been struggeling with the Backlight threshold feature.
I set a value, say "20" and the black light becomes greyish WHICH IS OK!

According to the explanation however this setting should be deactivated during effects and when status is "OFF".

But when I set the effect Knight rider for example the grey stays on.
Also when I turn my tv off it stays on

Please fix this issue because I really love this project!

PR #682, opened 2020-02-15
fix: Backlight/BBDetector during effects

brindosch:
- fix: Backlight/BBDetector during effects

Fixes: #674

### DIFF
--- a/include/hyperion/Hyperion.h
+++ b/include/hyperion/Hyperion.h
@@ -458,11 +458,10 @@ public slots:
 private slots:
 
 	///
-	///	@brief Apply ComponentRegister emits for COMP_ALL. Enables/Disables core timers
-	///	@param comp   The component
-	///	@param state  The new state of the component
+	///	@brief Handle whenever the visible component changed
+	///	@param comp      The new component
 	///
-	void updatedComponentState(const hyperion::Components comp, const bool state);
+	void handleVisibleComponentChanged(const hyperion::Components& comp);
 
 	///
 	///	@brief Apply settings updates for LEDS and COLOR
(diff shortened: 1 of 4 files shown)


brindosch:
As you may noticed, you can link one or more issues to close them automatically when a PR is merged.
`Fixes: #674`
@Lord-Grey

T-IOOO:
Backlight threshold is now beeing deactivated during effects but still there are some issues remaining.

When i power off my tv the backlight stays on.

Then a weird issue for those of you running homebridge with homebridge-hyperion-light
if i switch ambilight off backlight stays on. If i quickliy play around with the switch then backlight switches off.

I have not tested this fully but it would be great to have a little bit of lightning during ”black screen moments”

brindosch:
> When i power off my tv the backlight stays on.

How do you "power off"?

>  If i quickliy play around with the switch

What action is executed with this "switch"?

T-IOOO:
By power OFF I mean the Receiver. I usually see the rainbow colors for like 10 seconds then the LEDs goes off totally. But when having backlight to a certain value it is going to stay on. The ”TV Ambilight” switch for homebridge is acting weird when power on/off. Backlight stays on if switch on and then off... but tap the switch two times fast will switch leds off. I think i might have a power backfeeding issue from my receiver since both the hdmi splitter and hdmi2av dont need any power supply connected to them 😬

brindosch:
> The ”TV Ambilight” switch for homebridge is acting weird when power on/off. Backlight stays on if switch on and then off

Which action does this switch perform on Hyperion side.

T-IOOO:
This switch makes leds either go on or off. I am not sure how it works. There is another separate button for ”tv backlight”,

The backlight threshold stays on when switch is in OFF mode. tapping quickly twice (on-off-on-off) wil turn leds off completely.

I dont think homebridge is the problem itself. Maybe it is my power backfeed. I dont know how hyperion recognize OFF?

PR #709, opened 2020-02-26
Fixes for multiple memory leaks and segfault issues in Flatbuffer Forwarder

### LINKED ISSUE #652: Forwarding via flat-buffers not working.
I have NG built from a recent commit #4c0c102 (10/11/2019) on a Vero4K (OSMC) and an RPi3 (RaspbianLite)

Platform capture works on both devices and can be seen in the webGUI live preview.

On the Vero4K I have "forwarder" activated, with the "JSON target" and "flatbuffer target" set to the IP address of the RPi3 and the default ports.
Debug output is:
[hyperiond JSONSERVER] <INFO> Started on port 19444
[hyperiond FLATBUFCONNECTION] <INFO> Connecting to Hyperion: 192.168.1.250:19401
[hyperiond NETFORWARDER] <INFO> Forward now to json targets '192.168.1.250:19446'
[hyperiond NETFORWARDER] <INFO> Forward now to flatbuffer targets '192.168.1.250:19401'​

On the RPi3 I have "Flatbuffers Server" and "Protocol Buffers Server" activated, set to the same ports as above.
Debug output is:
[hyperiond JSONSERVER] <INFO> Started on port 19444
[hyperiond FLATBUFSERVER] <INFO> Started on port 19401
[hyperiond PROTOSERVER] <INFO> Started on port 19446​

So how should I see the Vero4K forwarded onto the RPi3?
I assumed there would be an option in the RPI3 "remote control" to select the Vero4K as a source, but I only have the usual:
System Platform Capture: (Dispmanx)
System Effect: Warm mood blobs​

It is unclear if this is a bug or operator error.
Is there something else I should be doing?

Many Thanks,

Filoni:
**Summary**
Flatbuffer forwarder was not working correctly, and crashing Hyperion.

The issue turned out to be caused by incorrectly assuming that the socket buffer would only contain one header and one flatbuffer message, resulting in a race condition which could cause the flatbuffer library to fail to verify the corrupted buffer it was passed. 

Upon failing to verify the buffer, the client attempts to sendErrorReply, which due to the same issue on the server causes Hyperion to crash with "Unable to parse reply".

Additionally, after fixing this issue, I discovered there was a memory leak caused by sendMessage on the server only conditionally clearing the flatbuffer builder. If a client was not connected, this caused the builder to grow without bound until it causes a segfault/badalloc. I fixed this by clearing the builder outside of sendMessage wherever it is used, and edited the client class to be consistent.

I believe this PR fixes #652 

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/libsrc/flatbufserver/FlatBufferClient.cpp
+++ b/libsrc/flatbufserver/FlatBufferClient.cpp
@@ -44,7 +44,7 @@ void FlatBufferClient::readyRead()
 		if((uint32_t) _receiveBuffer.size() < messageSize + 4) return;
 
 		// extract message only and remove header + msg from buffer :: QByteArray::remove() does not return the removed data
-		const QByteArray msg = _receiveBuffer.right(messageSize);
+		const QByteArray msg = _receiveBuffer.mid(4, messageSize);
 		_receiveBuffer.remove(0, messageSize + 4);
 
 		const auto* msgData = reinterpret_cast<const uint8_t*>(msg.constData());
@@ -113,6 +113,8 @@ void FlatBufferClient::registationRequired(const int priority)
 
 		// send reply
 		sendMessage();
+
+		_builder.Clear();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


brindosch:
Thank you for your contribution. Looks good!

Issue #927, opened 2020-08-01
HDR tone mapping and FPS software limit feature... partially implemented.


awawa-dev:
[final_lut_fps_limit_2020_08_01_patch.diff.txt](https://github.com/hyperion-project/hyperion.ng/files/5011116/final_lut_fps_limit_2020_08_01_patch.diff.txt)
[lut_rgb.zip](https://github.com/hyperion-project/hyperion.ng/files/5011120/lut_rgb.zip)

#### What problem does this feature solve?
Using most of the video grabbers or matrix/scallers with HDR content causes color degradation (bleak colors).
We can't repair the hardware but we can somehow corrected it using 3D LUT HDR to SDR tone mapping.

HDR test file: "HDR10-Sony-Bravia-OLED demo" (before and after)
![03](https://user-images.githubusercontent.com/69086569/89152714-f2c9a000-d563-11ea-8f69-8f00098f868f.jpg)
![03-corrected](https://user-images.githubusercontent.com/69086569/89152730-f9f0ae00-d563-11ea-84d2-c59f227d8cb3.jpg)
![07](https://user-images.githubusercontent.com/69086569/89152740-fc530800-d563-11ea-90f5-9d0dfc4ab6df.jpg)
![07-corrected](https://user-images.githubusercontent.com/69086569/89152749-fe1ccb80-d563-11ea-991d-9c11a18d92f4.jpg)
![08](https://user-images.githubusercontent.com/69086569/89152754-ffe68f00-d563-11ea-9474-46809db4bb1d.jpg)
![08-corrected](https://user-images.githubusercontent.com/69086569/89152760-0117bc00-d564-11ea-8799-eef919400a9d.jpg)
![12](https://user-images.githubusercontent.com/69086569/89152765-0248e900-d564-11ea-958c-63aee2046461.jpg)
![12-corrected](https://user-images.githubusercontent.com/69086569/89152772-037a1600-d564-11ea-967b-36dd63086655.jpg)
![25](https://user-images.githubusercontent.com/69086569/89152778-04ab4300-d564-11ea-8c79-4d5382cafaad.jpg)
![25-corrected](https://user-images.githubusercontent.com/69086569/89152787-06750680-d564-11ea-81fe-dd76d7cc9588.jpg)
![28](https://user-images.githubusercontent.com/69086569/89152800-0b39ba80-d564-11ea-85a9-9b6abd0ebc4a.jpg)
[...]

Issue #524, opened 2018-12-07
Hyperion HDR


malalam:
Is there a proper way to get HDR mapping? I use a splitter from my AVR but when viewing HDR content, hyperion doesn't know how to map the extended colorspace.

RickDB:
Problem is getting Hyperion to tell playback contains HDR metadata so we need the players to send it for instance using a new API command, right now I solve this problem like this:

AVR -> HDFury Integral 2 -> HDMI to CVBS -> CVBS USB - > Hyperion

It's a long chain but does get the job done:

What the HDFury can do is convert it to 1080P @ 8bit with colors pace correction, this means both HDR and non-HDR display display the right colors.

With 10/12bit LEDs in theory we can do a 1:1 colorspace mapping for HDR however either Hyperion or any Arduino device in between would need support for it.

snikcers:
@RickDB Hi I do not get the right colorspace mapping with the Integral2. Can you give me your settings?

Greetings

Lucas

oynek:
Would really appreciate this too. This is the only reason I hesitate to buy all that stuff.

TechnoHoschi:
old Topic, i know ;)

vertex und integral2 cannot do hdr tone mapping!, only with the x4 is this possible

[](https://hyperion-project.org/threads/4k-hdr-and-washout-fixing.3685/)
[](https://hyperion-project.org/threads/hyperion-hdr-sdr-grabber-problematik.3690/#post-16172)

Issue #497, opened 2018-01-26
HDR Feature


Iridahonda:
This is more of a feature request:

When watching HDR video the Hyperion colours are washed out due to not be able to handle BT 2100 video properly in the grabber.  One potential solution could be to add a remote control button that increases the colour vibrance by a settable amount so HDR video can be watched in a good way.

salondigital:
Hi!

What file I need to edit for this? V4L2Grabber.cpp?

Thanks!

penfold42:
What pixel formats does your grabber support ?
Would it be easier to do this via v4l control settings ?

Iridahonda:
Hi

It doesn’t support HDR but I am thinking this is what I need:

https://hyperion-project.org/threads/color-correction-add-3d-lut-correction.2594/

I see lildadou has made a start with a port to Hyperion.

Br

/T

PR #928, opened 2020-08-01
Hdr tone mapping (for all input color encoding) & frame skipping for saving resources

awawa-dev:
Fixes: #927
also affected: #497, #524
------------------------------------------------------------------------------------------------------------------------------------------
3D LUT HDR to SDR tone mapping based on the solution from fast-hdr project. RGB LUT file (lut_rgb.3d) must be present in the Hyperion configuration folder. There will be a HTML5 page that could generate that file or user can always download prebuild one: [lut_rgb.zip](https://github.com/hyperion-project/hyperion.ng/files/5011243/lut_rgb.zip)

Some theory:

1) after analyzing solution from fast-hdr its seems, that modern 1080p/4k capable grabbers and scallers produce SDR output from HDR input that has broken gamma part. The chrominance is quite well preserved to my surprise .
The output gamma has characteristic of SMPTE ST 2084 HDR transformation. We must reverse it to XYZ coordinates and apply some simply linear gain (for ex. L=50).
eq

2) apply one of the tone mapping algorithm (hable,  Reinhard) or leave it alone

3) revert back gamma from XYZ to sRGB
Clipboard02

The process is quite challenging for CPU so no surprise that devices cant handle it. LUT table solves that problem, but to be fully efficient must be implemented on the hardware. This patch also produces quite good results but there is one limitation: we don't have some information that are encoded in HDMI protocol meta-data.
 
---------------------------------------------------------------------------------------------------------------------------------------------

There is also new settings that allows skipping frames to save resources.

new_front

test3

[](https://user-images.githubusercontent.com/69086569/90320238-b02e9d00-df3f-11ea-891e-c428620d6716.jpg)

---------------------------------------------------------------------------------------------------------------------------

**Frame skipping for saving resources:**

current status: 800x600 30fps, decimation = 1 => CPU-core reaches it's limit
perf800600_30

current status: 800x600 custom 5fps, decimation = 1 => no change, CPU-core reaches it's limit, hardware FPS limit overrides custom one
perf800600_5

new feature: 800x600 new software frame skipping 5 for 30FPS so 6fps, decimation = 1 => CPU-core usage lowered to 60%
perf800600_skipping6

- [x ] Feature

- [x ] No


awawa-dev:
I think I can easily expand that solution to other color spaces & encoding (beside MJPEG) in ImageResampler. But first we should review & test that patch.

tihoangyeudau:
nice, hope it will work well

m-seker:
Why don't we just add ````lut_rgb.3d```` as a resource ?

libsrc/grabber/v4l2/V4L2Grabber.cpp
```diff
@@ -71,11 +90,66 @@ V4L2Grabber::V4L2Grabber(const QString & device
 	setInput(input);
 	setWidthHeight(width, height);
 	setFramerate(fps);
-	setDeviceVideoStandard(device, videoStandard);
+	setDeviceVideoStandard(device, videoStandard);	
+}
+
+void V4L2Grabber::loadLutFile(QString path)
```
m-seker, line 96:
```
QFile file (path +"/lut_rgb.3d");
QByteArray buffer;
if (file.open(QIODevice::ReadOnly))
{
    buffer = file.readAll();
}
file.close();

// use buffer...
```

config/hyperion.config.json.commented
```diff
@@ -138,7 +140,9 @@
 		"sDVOffsetMin"         : 0.25,
 		"sDHOffsetMin"         : 0.25,
 		"sDVOffsetMax"         : 0.75,
-		"sDHOffsetMax"         : 0.75
+		"sDHOffsetMax"         : 0.75,
+		"hdrToneMapping"		   : false,
+		"fpsSoftwareDecimation"	: 1
```
m-seker, line 145: This would be more clear if we name it capture rate or capture frequency and use accordingly, like the other grabbers.

libsrc/grabber/v4l2/V4L2Grabber.cpp
```diff
@@ -1013,6 +1087,10 @@ int V4L2Grabber::read_frame()
 
 bool V4L2Grabber::process_image(const void *p, int size)
 {
+
+	if ( (_fpsSoftwareDecimation > 1) && ((_currentFrame++) % _fpsSoftwareDecimation == 0))
```
m-seker, line 1091: Why do we need this when we can change capture frequency ?

awawa-dev:
> 
> 
> Why don't we just add `lut_rgb.3d` as a resource ?

It's ~50MB raw though it compresses really well. As an alternative I'm thinking about WWW subpage in Hyperion panel that could generate that file dynamically in HTML5: with more option to tune that 3d lut table by user. Then he would have to upload that generated file to config directory manually.

libsrc/grabber/v4l2/V4L2Grabber.cpp
awawa-dev: FPS settings from the device configuration is insufficient for high quality 1080/4k grabbers due to lack of resources on some configuration (Rpi3 in my case).

As for ezcap 269 https://postimg.cc/F1sgFK3N there is only a choice between 30 and 60fps: nothing between, nothing less. That causes some delay between movie on the TV and colors on the leds. Frame decimation somehow reduces that problem but works bad with mjpeg encoding in lower resolution (jpeg artifact have quite impact). Rejecting certain frames before processing them allows us to use higher resolutions or abandon frame decimation or to have less impact on CPU. All that reduces delay TV<->LED if that occurs for a price of lower FPS for Hyperion.

PS: USB 2.0 with mjpeg  bus is not a problem even on Rpi3 https://postimg.cc/4Yh1CFtc with 30 or 60 fps. But the CPU usage are over 100% mostly for only one core. Only 640x480 dec =1 and 800x600 dec =2 are acceptable with 30 FPS. Reducing FPS and CPU usage solved the lag problem in my case.

m-seker: I didn't know that frame rate setting is restricted to predefined values in some devices. 
```

                // Check if the device is able to accept a capture framerate set.
                if (streamparms.parm.capture.capability == V4L2_CAP_TIMEPERFRAME)
                {
                        streamparms.parm.capture.timeperframe.numerator = 1;
                        streamparms.parm.capture.timeperframe.denominator = _fps;
                        (-1 == xioctl(VIDIOC_S_PARM, &streamparms))
                        ?       Debug(_log, "Frame rate settings not supported.")
                        :       Debug(_log, "Set framerate to %d fps", streamparms.parm.capture.timeperframe.denominator);
                }
```
Does this also apply to resolution settings ? I mean, isn't it possible to reduce the resolution to something lower than 640x480 ? Because those values are higher than what we provide as default in other grabbers (Picture decimation is 8 by default, so a FHD image becomes 240x135). 

Or you tried and didn't like the results with lower resolution ? If so, what difference did it make to use a higher resolution ?

awawa-dev: I have attached screenshots:
800x600 30fps decimation = 1 => CPU-core reaches it's limit
perf800600_30

800x600 custom 5fps decimation = 1 => no change CPU-core reaches it's limit, hardware FPS limit override it
perf800600_3

and 800x600 new software frame skipping(5 for 30FPS) so 6fps, decimation = 1 => CPU-core usage lowered to 60%
perf800600_skipping6

But I've discovered that when I have added that option to the global config (for my use I hardcoded it) I made a school mistake in modulo formula for frame skipping :anguished: : 
IN PULL REQUEST:
`if ( (_fpsSoftwareDecimation > 1) && ((_currentFrame++) % _fpsSoftwareDecimation == 0))
+		return false;`
SHOULD BE OF COURSE:
`if ( (_fpsSoftwareDecimation > 1) && ((_currentFrame++) % _fpsSoftwareDecimation **!=** 0))
+		return false;`
otherwise it does skipping but not I mean to.

marcosvrs:
Why won’t you separate the js code in a file itself? (Don’t know if you are still developing it or not)

awawa-dev:
> 
> 
> Why won’t you separate the js code in a file itself? (Don’t know if you are still developing it or not)

it's already present in the latest commit

marcosvrs:
> > Why won’t you separate the js code in a file itself? (Don’t know if you are still developing it or not)
> 
> it's already present in the latest commit as one page you can open using link in the new option's description. It's all in one, standalone solution as I planned without any external dependencies. Want to separate it from the hyperion portal because don't want to mess with it's design as it's only an option that not everyone needs it. I need full page width to compare before and after result for the test image also.
> 
> For now I think I'm done with changes. I'll investigate QByteArray to see if there will be any negative performance impact vs using unsigned char* buffer.

It won’t affect the design of the Hyperion portal and will look more cleaner if you segregate the JavaScript part of your HTML and call the file only on this specific HTML page
Like ‘<script src=“/js/content_lut_generator.js”>’.
Even some of the other pages has it’s own js files which are requested only on its needed page.

The code to generate the file via js is already complex, and adding it to the HTML file doesn’t help to understand it, IMO

(PS: I don‘t know the current code guideline of this project. the suggestions that I’m giving are based only on my dev experience.)

awawa-dev:
Yes I know, but there is to much integration with UI. As it's only one totally separate page & without any layout it don't think it's worth effort to rewrite and separate view & controller on this stage before test.

libsrc/utils/ImageResampler.cpp
```diff
@@ -150,3 +271,28 @@ void ImageResampler::yuv2rgb(uint8_t y, uint8_t u, uint8_t v, uint8_t &r, uint8_
 	g = clamp((298 * c - 100 * d - 208 * e + 128) >> 8);
 	b = clamp((298 * c + 516 * d + 128) >> 8);
 }
+
+
+void ImageResampler::yuv2rgbHDR2SDR(uint8_t y, uint8_t u, uint8_t v, uint8_t &r, uint8_t &g, uint8_t &b, unsigned char *lutBuffer)
```
m-seker, line 276: There is another yuv conversion function in this class, I assume you just followed it but in the master I changed it. There is already a color conversion utility file called ColorSys and it makes more sense to keep all conversion code there. Can you move this one to ColorSys as well ? Clamp is already there.

libsrc/grabber/v4l2/V4L2Grabber.cpp
```diff
@@ -71,11 +80,66 @@ V4L2Grabber::V4L2Grabber(const QString & device
...
+	_hdrToneMappingEnabled = enable;
+	if (lutBuffer!=NULL || !enable)
+		Debug(_log,"setHdrToneMappingEnabled to: %s", enable ? "true" : "false");
+	else
+		Error(_log,"ERROR! setHdrToneMappingEnabled was to: enable, but the 3DLUT file (lut_rgb.3d)  was missing in the configuration folder");		
 }
 
 V4L2Grabber::~V4L2Grabber()
 {
+	if (lutBuffer!=NULL)
```
m-seker, line 139: no need for check, free(nullptr) is safe

libsrc/grabber/v4l2/V4L2Grabber.cpp
```diff
@@ -71,11 +80,66 @@ V4L2Grabber::V4L2Grabber(const QString & device
...
+	Debug(_log,"setFpsSoftwareDecimation to: %i", decimation);
+}
+
+void V4L2Grabber::setHdrToneMappingEnabled(bool enable)
+{
+	_hdrToneMappingEnabled = enable;
+	if (lutBuffer!=NULL || !enable)
+		Debug(_log,"setHdrToneMappingEnabled to: %s", enable ? "true" : "false");
+	else
+		Error(_log,"ERROR! setHdrToneMappingEnabled was to: enable, but the 3DLUT file (lut_rgb.3d)  was missing in the configuration folder");
```
m-seker, line 134: Error already prints that it is an error, no need for "ERROR: ..." :)

libsrc/utils/ImageResampler.cpp
awawa-dev: I revert back my previous attempt to merge as I saw them. OK, I'll move that part of code to ColorSys to unify changes.

libsrc/grabber/v4l2/V4L2Grabber.cpp
awawa-dev: ...but you can't miss that second upper case ERROR in the debug log when something goes wrong ;) OK, I remove it.

libsrc/grabber/v4l2/V4L2Grabber.cpp
awawa-dev: ok

UniCizin:
nice work, i just compiled your fork and will test it a bit.
any hints on generating a good lut file?

awawa-dev:
> 
> 
> nice work, i just compiled your fork and will test it a bit.
> any hints on generating a good lut file?

Default settings are prepared for my ezcap269 (and as I know it fit at least one other device).
I have tested them with some samples found in the internet related to HDR color washed out problem and them passed that tests also.
You should start with gain 100 (maybe a little too bright for hable) and check if the result is comparable to TV (or HDR capable video player on PC)

1) take few test screenshots from your grabber in LED visualization [(for example)](https://hyperion-project.org/attachments/libreelec-png.1990/). Crop then with irfanview if needed.
2) run the creator 
link
and upload test image, then apply LUT and compare to the reference image on TV or PC

domeOo:
What is the reason that this isn’t merged? @awawa-dev

S0und:
@domeOo my guess is that they rather released they own version of hyperion.ng with the HDR correction in it, which is a same. 

https://github.com/awawa-dev/HyperHDR

PR #1043, opened 2020-10-12
Feature: Temporal Color Smoothing with variable decay-rate for long period average windows

The-Master777:
**Summary**

In addition to the linear interpolation of LED colors that was previously supported by LinearColorSmoothing, a new algorithm - named **"Decay"** - is made available to the user. The _Decay_ algorithm extents the color smoothing beyond the existing limitations of the linear interpolation. 

**This incorporates support for:**
- long smoothing times of multiple seconds
- microsecond precision multi-frame smoothing
- high input frame-rates ≥ 60 fps
- very high output frame-rates ≥ 240 fps
- linear decay: similar to the previous LinearColorSmoothing
- exponential decay: non-linear weighting of input frames for nice _lightbulb decay_ effect
- temporal dithering for best color quality

For detailed description see *Other Information* below.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

Options for configuring the decay-parameters are made available to the user. Please note the changes made to `schema-smoothing.json`.

- [ ] Yes
- [x] No

- [x] A convincing reason for adding this feature: The existing linear color smoothing algorithm was only capable of interpolation between the previous and the target colors. For long smoothing windows or high frame rate the results were not sufficient.

**Other information:**

The temporal smoothing now supports very long settling-times (multiple seconds) that allows frame interpolation to span across several frames of LED colors. For calculating intermediate frames it uses a decay-based algorithm that interpolates colors based on the age of previous frames and a given decay-power.

For each moment, all ingress frames that were received during the smoothing window are reduced to the concrete color values using a **weighted moving average**. This is done by applying a decay-controlled weighting-function to individual the colors of each frame.

As the individual intermediate frames are solely dependent on the received frames – **not** on previous interpolated values – **very high frame-rate** targets (>>120 fps) can be met without noticeable loss of precision. The _Decay_ algorithm effectively implements a conversion of the ingress frame-rate to the configured egress rate by generating high precision intermediate frames for every moment in time.

**Decay Value**

The decay-power influences the weight of individual frames based on their 'age'.

- A decay value of 1 indicates linear decay. The colors are given by the moving average with a weight that is strictly proportionate to the fraction of time each frame was visible during the smoothing window. As a result, equidistant frames will have an equal share when calculating an intermediate frame.

- A decay value greater than 1 indicates non-linear exponential decay. With higher powers, the decay is stronger. I.e. newer frames in the smoothing window will have more influence on colors of intermediate frames than older ones. This leads to a _lightbulb_ effect, where the ramp-up response is fast but the decay is slower.

Typical non-linear decay values are are less than 8, but a maximum decay-power of 20 is supported for experimental use with extreme settlingTimes > 5s. 

**Dithering**

A temporal dithering algorithm is used to minimize rounding errors, when downsampling the average color values to the 8-bit RGB resolution of the LED-device. Effectively, this performs diffusion of the residual errors across multiple egress frames. Users will perceive accurate colors when high output framerates > 120 fps are used.


Paulchen-Panther:
@The-Master777 Ready to merge?

The-Master777:
> @The-Master777 Ready to merge?

From my side its ready. 

Please check the changes made to `schema-smoothing.json` are fine for you as well, as these are facing the user.

Lord-Grey:
@ManusSs Hi, I already merged your change, but currently I am looking to get away from compiler warnings...

Would you mind addressing the compiler Warning for
`LinearColorSmoothing.cpp(273): warning C4293: '<<': shift count negative or too big, undefined behavior`
on
`const uint64_t scale = (1l<<FPShift) * static_cast<double>(weight);`

Thank you!

The-Master777:
Sure, i'll give this a look. On which platform do you observe this behaviour? 

The idea is, to use fixed point arithmetics on Int64 instead of using the FPU for aggregating the Color components as this seems to be more efficient.

I assume (1 << FPShift) *weight == Pow(2,FPShift) * weight, where `FPShift = (sizeof(uint64_t)*8 - (12 + 9))` = 64  - 21 = 43 bits. So I basically want to multiply 2^43 with the weight.

Maybe MSVC is somehow different on this compared to GCC? 
--> https://docs.microsoft.com/de-de/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4293?view=msvc-160

So might `const uint64_t scale = (static_cast<unsigned __int64>(1l)<<FPShift) * static_cast<double>(weight);´fix this?

Can you test on MSVC?

Lord-Grey:
Thanks for coming back quickly!

Indeed the Windows compiler brought-up this warning and with the __int64 cast it went away.
Nevertheless, __int64 does not work with gcc.

I changed the cast to 
`const uint64_t scale = (static_cast<uint64_t>(1l) << FPShift) * static_cast<double>(weight);`
which compiles on both environments.
As I am not into the depths or the logic. Is this fine with you?

The-Master777:
> Thanks for coming back quickly!
> 
> Indeed the Windows compiler brought-up this warning and with the __int64 cast it went away.
> Nevertheless, __int64 does not work with gcc.
> 
> I changed the cast to
> `const uint64_t scale = (static_cast<uint64_t>(1l) << FPShift) * static_cast<double>(weight);`
> which compiles on both environments.
> As I am not into the depths or the logic. Is this fine with you?

Thanks for going ahead. This is fine for me and works on my build aswell.

PR #1163, opened 2021-01-21
Custom Effects - Clean-ups and Enhancements

Lord-Grey:
**Summary**

Custom Effects - Clean-ups and Enhancements

- Fix: Custom Effect Templates (schemas) are now loaded
- Enhancement: Support Custom Effect Templates in UI for custom effect creation and configuration
- Code simplification & clean-up
- Fix: WebSockets, handling of fragmented frames
- Fix : Uploaded images were not found executing custom image effects

- [X] Bugfix
- [X] Feature
- [X] Code style update
- [X] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

New User Effect Template Support
image

User Templates are to be stored under "custom-effects/schema"

image

- [ ] Yes
- [X] No

- [X] A convincing reason for adding this feature

**Other information:**

This PR is based on and replaces PR #1158


EliteScientist:
@Lord-Grey The system effects are no longer there.

Using the Effect Configurator I can see the UI for my custom effect but when I first began to test it I was getting an ALSA error which makes me think the proper values weren't sent to the effect:

2021-01-22T04:35:47.781Z [hyperiond EFFECTENGINE] (ERROR) ###### PYTHON EXCEPTION ######
 2021-01-22T04:35:47.782Z [hyperiond EFFECTENGINE] (ERROR) ## In effect 'VU Meter'
 2021-01-22T04:35:47.782Z [hyperiond EFFECTENGINE] (ERROR) ## ALSAAudioError: No such file or directory [hw:1]
 2021-01-22T04:35:47.843Z [hyperiond EFFECTENGINE] (ERROR) ## Traceback (most recent call last):
 2021-01-22T04:35:47.843Z [hyperiond EFFECTENGINE] (ERROR) ## File "", line 18, in 
 2021-01-22T04:35:47.843Z [hyperiond EFFECTENGINE] (ERROR) ## alsaaudio.ALSAAudioError: No such file or directory [hw:1]
 2021-01-22T04:35:47.843Z [hyperiond EFFECTENGINE] (ERROR) ###### EXCEPTION END ######
 
its trying to get device 1 hw:1 and it defaults to device 3. After playing with it for a while then it started sending the correct values and I couldn't get it to error anymore.


image

libsrc/effectengine/EffectFileHandler.cpp
```diff
@@ -101,17 +114,14 @@ QString EffectFileHandler::saveEffect(const QJsonObject& message)
 	QString resultMsg;
 	if (!message["args"].toObject().isEmpty())
 	{
-		QString scriptName;
-		(message["script"].toString().mid(0, 1)  == ":" )
-			? scriptName = ":/effects//" + message["script"].toString().mid(1)
-			: scriptName = message["script"].toString();
+		QString scriptName = message["script"].toString();
```
EliteScientist, line 117: I take it that internal effects are handled differently?

EliteScientist: After looking at this code. I copied the effects files from the source folder to "/usr/local/share/hyperion/bin/effects" and now they show up. What do you think about using an environment variable like $HYPERION_HOME that by default would point to the hyperion folder: "/usr/local/share/hyperion/" then from there you have access to everything hyperion.

Lord-Grey: > It appears that internal effects went from ":/effects/" to "./effects/"

I had initially a different motivation to move away from the colon that an available JS routine could be used.
This nevertheless, did not paly out well, when Windows filenames had colons too.
Plus I might have not understood side effects which do not occur, if you just run in a dev-environment...

Long story short, I reverted to ":/effects" and I works now for me on Ubuntu and Windows.
This should then also remove the need to introduce new environment variables, as the relative addressing works.

Would you mind checking on your end again, please?

Lord-Grey:
> its trying to get device 1 hw:1 and it defaults to device 3. After playing with it for a while then it started sending the correct values and I couldn't get it to error anymore.

If there are still problem, share your custom  JSONs and py files with me.
Maybe with another pair of eyes, we get this sorted.

libsrc/effectengine/EffectFileHandler.cpp
EliteScientist: Awesome this is much better. The only thing I see is the indentation in the UI isn't there.

image

I want to submit this effect to the community also. There is one issue where I had to move some files around manually on the file system and compile hyperion with the version of python on the system.

I needed to install "pyalsaaudio" and its dependencies. Installing them into the OS Hyperion didn't pick them up. I had to copy the libraries I wanted into hyperion's lib/python folder.

Looking at some of the stuff here: https://stackoverflow.com/questions/2812520/dealing-with-multiple-python-versions-and-pip

It appears that if the python executable is included and configured to hyperion's folders then we may be able to install a pip and packages into it. This will give devs access to cool math libraries like numpy also.

The other thing would be to enable it to use the OS's python. this should enable it to pick up installed python libraries.

Lord-Grey: > The only thing I see is the indentation in the UI isn't there.

Indentions are back....

Lord-Grey: > The other thing would be to enable it to use the OS's python. this should enable it to pick up installed python libraries.

The current challenge why Hyperion are delivered incl. libraries is
a) There are systems which do not allow additional installations and are still on python 2, e.g. LibreElec, CoreElec
b) There had been some incompatibilities before

Given that the DIY aspect in this project is on the LEDs etc., most users are even overwhelmed accessing their box even via ssh. Not to say that navigating/installing under linux is even a bigger issue....

That is the rationale why hyperion is as much packaged as such.

Nevertheless, I like the audio effect idea, but rather spending the effort opening up to be flex on python, pip etc, I would suggest using your experience adding audio grabbing as a native feature into the platform.

EliteScientist: Awesome. I agree.

The only place I can see needing python packages is if any non-audio effects want to use numpy for mathematical calculations or something else.

Issue #1092, opened 2020-11-12
Color correction for grabbed picture


todely:
## Feature request
Add colors RGB, gamma, brightness and saturation correction to applied on capturing hardware flow to avoid bad / not exact color transcription from hyperion to led device

#### What problem does this feature solve?

Solve issue for grabber with bad / incorrect transcode signal and perhaps HDR signal.

#### What does the proposed API look like?

Adding parameters in capture hardware section or in picture processing section

#### How should this be implemented in your opinion?

By adding filter on flow coming from usb capture device or when hyperion analyse picture ?

#### Are you willing to work on this yourself?

Don’t have programming capabilities sorry

Lord-Grey:
Addressed by #1477

PR #1477, opened 2022-06-11
Saturation & Brightness/Value Gain using Oklab color space

### LINKED ISSUE #822: Saturation
## Feature request

 can't ask on the forum 

#### What problem does this feature solve?
Hey, it would be very great if the old saturation setting could be implemented. I personally liked this setting and missed it 
#### How should this be implemented in your opinion?
Increase the color so light red is red 
#### Are you willing to work on this yourself?
No, I can't. Sorry :(

### LINKED ISSUE #1142: Value & saturation gain and threshold
## Feature request
The old hyperion had a hsv, saturationGain and valueGain
And it also had a threshold that could be set.
The configuration of the old hyperion looked like this:
```
		"transform" :
		[
			{
				"id"   : "default",
				"leds" : "*",
				"hsv" :
				{
					"saturationGain"	: 1.2000,
					"valueGain"		: 1.1000
				},
				"red" :
				{
					"threshold" 	: 0.0200,
					"gamma"     	: 1.6000
				},
				"green" :
				{
					"threshold" 	: 0.0200,
					"gamma"     	: 1.6000
				},
				"blue" :
				{
					"threshold" 	: 0.0200,
					"gamma"     	: 1.6000
				}
			}
		],
```
The new hyperion ng seems to be missing the valuegain, saturationgain and threshold.
I would like to also see those options in the new hyperion ng.
I noticed that the philips hue Led device does have a valuegain setting under configuration --> LED hardware --> LED controller.
I basically want that option available for all platforms under configuration --> Image Processing.

#### What problem does this feature solve?
Specifically for my secondary lights throught the room (so not the LEDs behind the TV but other lights controlled through UDP raw) I would like to have a desent valuegain to have them most of the time turned on at full brightness and only change color but not brightness to still have some light in the room. Only at verry very dark scenes in a movie they schould start to dim. This was possible with the ValueGain in the old hyperion.

Besides these lights use a pretty big area of the TV to derive the color from, that is good to get a better average of the screen color, but it also makes them tend to white values a lot instead of nice saturated colors. Therefore the SaturationGain is also very handy for such lights.

At last the LEDs behind my TV tend to flicker a bit when at very low brightness, I previously used the threshold option of the old hyperion to turn LEDs off when the brightness was below 2% brightness.
I would very much like to have this option back under configuration --> Image Processing in the new hyperion NG.

#### How should this be implemented in your opinion?
The same way it was done in the old hyperion.

#### Are you willing to work on this yourself?
I actually am working on quite some other github projects, mainly HomeAssisistant related and written in python.
I am not fimiliar with the Hyperion code and it is written in c which I am less proficiant in.
Therefore I would love it if someone else could implement this that is already fimiliar with Hyperion.
I could then help other people with there issues on other pyhon github projects.

xkns:
**Summary**

Adds Saturation & Value Gain settings to the UI under `Image Processing` and adjusts the colors accordingly using [Okhsv](https://bottosson.github.io/posts/oklab/). As discussed on Discord this was done using a new `OkhsvTransform` class.

Fixes #822, #1092 and partially fixes #1142.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

[Images](https://imgur.com/a/HT7MIG3)

- [ ] Yes
- [x] No

- [x] A convincing reason for adding this feature

**Other information:**


xkns:
> This pull request **introduces 5 alerts** when merging [b874453](https://github.com/hyperion-project/hyperion.ng/commit/b8744535557117ebc9ace83fb9bb1dd6202451d6) into [359618a](https://github.com/hyperion-project/hyperion.ng/commit/359618a4eefd4fc4a27d49993d9960e8f896c544) - [view on LGTM.com](https://lgtm.com/projects/g/hyperion-project/hyperion.ng/rev/pr-de484607add39959ac3de47b66f8d5cbe502d5bf)
> 
> **new alerts:**
> 
>     * 4 for Multiplication result converted to larger type
> 
>     * 1 for Declaration hides parameter

These are all in the Oklab reference implementation. Not sure how easy it is license wise to try to fix them.

cjohnsto-nz:
I deeply hope that this PR is accepted. I built this and the saturation and brightness gain controls put Hyperion over the line for me.

Would the translation to OKLAB facilitate the implementation of ICC profiles for calibration? I find that my particular LEDs render blue midtones as a rich cyan in a way that cannot be corrected through Hyperion's calibration options without ruining the white balance.

xkns:
> I deeply hope that this PR is accepted. I compiled this and the saturation and brightness gain controls put Hyperion over the line for me.

I've been in contact with the maintainer (@Lord-Grey) and I'm optimistic that this will get merged at some point. We'll see how the code review goes...

> Would the translation to OKLAB facilitate the implementation of ICC profiles for calibration? I find that my particular LEDs render blue midtones as a rich cyan in a way that cannot be corrected through Hyperion's calibration options without ruining the white balance.

We have also started to look into this issue. While I believe that it is theoretically possible to get a good color correction profile set using the current settings (color setpoints + gamma curves) we agree that it isn't very intuitive to do so.

I don't think that importing ICC files is the solution here as that would require users to use some external software to generate those files. Fine tuning would also be very annoying as one would have to import the newly adjusted ICC file every single testing iteration.

I'm sure that we can find a more user friendly way but that will be in a new PR. Whether or not Oklab will play a part in this is to be seen but the necessary facilities will be present when this PR is merged.

Lord-Grey:
@xkns Sorry, for only coming back on this now.

a) As I am not too detailed in the Color transformation subject, maybe you could help me in understanding why the saturation and value gain are transformed before the RGB color transformations. By heart I would have expected that first Colors are mapped and then gains applied. Maybe I am missing some theory here....

b) What is the number range for saturation and value gain?
Having 0-100 with a stepping of 0.1 does not seem to be reasonable. 
Or did you wanted to express a range of 0-100 percent?
(Some definitions in the JSONs are currently not ok, but I would like to update them depending on the outcome of the above)

c) Should the RGB+SatGain+ValGain an alternative way to the existing calibration rather then one on top?
Would a user be able to control 11+ values to achieve a good calibration?

d) What would be the guidance for a standard user going forward?
What are the elements that should be shown on the screen by default? (With input I will make it happen)
Via "expert" level all elements could be shown...

Sorry for those many questions, but I would like to get a better feels in which direction the target design should go.

Lord-Grey:
> Edit: As the reference implementation is licensed under MIT it appears that we can modify and redistribute it under any license as long as the copyright is kept. But I'm not a lawyer.

As this is MIT, we can modify the code. Good practice is to keep the original copyright, plus adding info on change
e.g.:

`// Copyright(c) 2021 Björn Ottosson + Updates by xkns`

xkns:
> a) As I am not too detailed in the Color transformation subject, maybe you could help me in understanding why the saturation and value gain are transformed before the RGB color transformations. By heart I would have expected that first Colors are mapped and then gains applied. Maybe I am missing some theory here....

The order of operations should be irrelevant for hue and saturation. If you were to first saturate a color and then map it or first map it and then saturate the mapped color shouldn't make a significant difference (some rounding errors might occur).
What does make a difference is when you are limiting the brightness through the existing mapper and were to then apply some brightness gain (>= 1.0). In that case you would exceed the maximum brightness as set by your mapping.

There are basically three options here:
1. Okhsv transform last: You would need to limit the max brightness again and by the same value as given for the other mapping.
2. Okhsv transform somewhere in the middle: The RgbTransform gets compiled into what is effectively a mapping matrix. This is very efficient as it can do all operations in one go. Splitting that matrix up into two in order to fit the Okhsv transform in the middle would lose some of that benefit.
3. Okhsv transform first: Let RgbTransform do its thing.

> b) What is the number range for saturation and value gain? Having 0-100 with a stepping of 0.1 does not seem to be reasonable. Or did you wanted to express a range of 0-100 percent? (Some definitions in the JSONs are currently not ok, but I would like to update them depending on the outcome of the above)

`1.0` is the neutral and (intended) default value for both settings. A brightness gain of `0.0` doesn't make too much sense to me as that would always result in total `0, 0, 0` blackness. Desaturation gain of `0.0` would completely desaturate leaving only grayscale colors. Values smaller than `0.0` have no valid result. On the opposite end gains of `255` and larger would probably always result in perfect white `255, 255, 255` although I haven't tested this extrema. `100` brightness gain is effectively the current brightness times 100. I wasn't sure exactly where to put the upper limit.

Should you want to use percent instead all values would simply have to be multiplied by 100 (including the upper limits).

Using other scales would also be possible but a simple multiplier is quite intuitive in my opinion.
Brightness gain for example could use the same input curve as gamma which would make them act identical given that all three gamma values are the same. You would however lose the ability to "overdrive" the brightness (which might not even be a bad thing).
I would be happy to implement something else instead if you have an idea.

> c) Should the RGB+SatGain+ValGain an alternative way to the existing calibration rather then one on top? Would a user be able to control 11+ values to achieve a good calibration?

ValGain is redundant in this sense because you should theoretically be able achieve the same effect by setting all three gamma values just right. What it can't do however is account for different physical light output by various color channels on the hardware. This is important for getting mixed colors (including gray tones) right.

I'm not sure if I would count SatGain as a calibration measure or simply an aesthetic effect. It's not one of the "calibration settings" you would find in other pieces of software but you could argue that (especially in HDR situations) the LED color output might be lacking in saturation / looking washed out. In that sense it would calibrate something that the other existing settings cannot account for.

> d) What would be the guidance for a standard user going forward? What are the elements that should be shown on the screen by default? (With input I will make it happen) Via "expert" level all elements could be shown...

This probably needs some additional discussion but I would say that saturation is something that a standard user has some intuition for. As it isn't really redundant there isn't really an alternative expert/non-expert setting to show instead.

Brightness gain on the other hand is mostly identical (the curves diverge more as you get farther away from the neutral value) to gamma settings if all gamma settings have the same value. So the gamma settings could be regarded as the expert brightness gain. This is especially the case if we were to change the input curve to match the gamma formula.

> Sorry for those many questions, but I would like to get a better feels in which direction the target design should go.

Thanks for taking the time and sorry for messing up the JSON definitions.

Lord-Grey:
@xkns Ready to merge from your perspective?

xkns:
> @xkns Ready to merge from your perspective?

Yes ready to merge

PR #1753, opened 2024-06-08
DDA grabber: Set a 500ms timeout when waiting for a new frame

davidsansome:
**Summary**

Previously the DDA grabber would wait forever (set an `INFINITE` timeout) for a new frame. The API only delivers a new frame to the application if some part of the display has changed, so if nothing on the display was changing then the `AcquireNextFrame` call would block for a long time, Hyperion wouldn't deliver new data to the LEDs, and WLED would shut off.

This change sets a 500ms timeout, which ensures the grabber returns a new image to Hyperion at least every 500ms even if nothing on the display has changed.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**


Lord-Grey:
Have you seen that WLED has shut off the lights when there is no update in a given timeframe.
I though in live streaming mode , which we are using, this would not happen. … but I might remember incorrectly.
In which  WLED version did it occur?

I get that the infinite blocking might cause certain issues, nevertheless, if an LED device is not stable showing an update or a longer period of time, the better approach would be to set a rewrite time at the device. That would avoid to go through full processing.

Therefore, I would ilke to understand, if the rationale is a „misbehavior“ at the LED side that is fixed at the grabber or if the infinite wait causes problems during frame capturing.

Thank you!

davidsansome:
Ah sorry for the late reply @Lord-Grey!

Yep I'm definitely seeing WLED shut off the lights if there's no update. I'm using version 0.15.0-b2. I agree it would be better to have a rewrite time at the device, but I wasn't sure if blocking the grabber thread indefinitely would cause other issues as well.

Issue #1956, opened 2025-12-27
RGW -> RGBW conversion algorithm to subtract white with a custom colour temperature


lougovsk:
#### What problem does this feature solve?

Currently Hyperion supports multiple methods to convert RGB to RGBW data. However, they are generic and do not take into account the exact white colour temperature specific LED strip can have. It would be great to have another RGB->RGBW conversion algorithm which would take colour temperature of white LED channel as input and subtract it from the RGB signal. 

#### What does the proposed API look like?

This won't require a new API. It might effect the configuration menus to provide another option when choosing for a conversion algorithm with an colour temperature parameter field.

#### How should this be implemented in your opinion?

It is actually a fairly straightforward. The change will mostly effect Rgb_To_Rgbw class. 

#### Are you willing to work on this yourself?
Yep. I have a prototype, which I can share.

PR #1993, opened 2026-03-30
Adding support of subtracting of a white with a given color temperature

lougovsk:
**Summary**

- [ ] Bugfix
- [X ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

After
Screenshot 2026-03-31 at 00 46 00

- [ ] Yes
- [X ] No

**Other information:**
This MR corresponds to the #1956


Lord-Grey:
@lougovsk Thank you for your contribution and PR issues. This is well received and appreciated.

I did some smaller updates for consistency with the overall project:

1. There is an existing KelvinToRgb function already in place based on the same algorithm.
I removed the duplicate code and updated the related code.

**Question:** Is there a specific reason why the Temperature default is 5000 and not 6600 which is the neutral one?
Is this e131 device related?

2. I updated the json to have it aligned with the standard that units are appended to the elements rather showing in the label.

3. I got the AI review feedback as following and pushed a proposed fix [1f6c1be](https://github.com/hyperion-project/hyperion.ng/pull/1993/commits/1f6c1beafbe43798723748e3704f8bf4777a7c4e).

The finding was as following

> The new white-channel scaling in [RgbToRgbw.cpp:150](vscode-file://vscode-app/snap/code/238/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [RgbToRgbw.cpp:151](vscode-file://vscode-app/snap/code/238/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is inconsistent with the rest of the RGBW model and breaks the main use case of this feature. The code derives the white output from ratio × sumW, then caps it at 255 and recomputes the subtraction ratio from that capped value. For common white temperatures this means even an exact chromaticity match does not collapse to the white channel. I checked the math directly: at 6500 K, input 255/255/255 produces white 255 but still leaves RGB 169/170/171; at 2000 K, an input equal to the computed white basis still leaves RGB 94/50/5. That means sub_ktemp_white cannot actually subtract the configured white cleanly before merge. The fix should scale the white channel as a 0..255 scalar of the chosen white basis, then subtract that same scalar fraction from RGB, instead of normalizing by the sum of the basis channels.

Maybe you can check, if it is in line with your intend.
If the fix does not make sense we can roll it back....

lougovsk:
> @lougovsk Thank you for your contribution and PR issues. This is well received and appreciated.
Hi @Lord-Grey  thanks a lot! And sorry for a delayed reply. 
> 
> I did some smaller updates for consistency with the overall project:
> 
> 1. There is an existing KelvinToRgb function already in place based on the same algorithm.
>    I removed the duplicate code and updated the related code.
Ah... Sorry, I didn't check the code base properly, so implemented the duplication. Really appreciate, that you've taken care of removing it.
> 
> **Question:** Is there a specific reason why the Temperature default is 5000 and not 6600 which is the neutral one? Is this e131 device related?
No specific reason. Just picked some mid range temperature as a default, not too warm, not too cold. Any reasonable default should be fine.
> 
> 2. I updated the json to have it aligned with the standard that units are appended to the elements rather showing in the label.
Thanks a lot!
> 3. I got the AI review feedback as following and pushed a proposed fix [1f6c1be](https://github.com/hyperion-project/hyperion.ng/pull/1993/commits/1f6c1beafbe43798723748e3704f8bf4777a7c4e).
> 
> The finding was as following
> 
> > The new white-channel scaling in RgbToRgbw.cpp:150 and RgbToRgbw.cpp:151 is inconsistent with the rest of the RGBW model and breaks the main use case of this feature. The code derives the white output from ratio × sumW, then caps it at 255 and recomputes the subtraction ratio from that capped value. For common white temperatures this means even an exact chromaticity match does not collapse to the white channel. I checked the math directly: at 6500 K, input 255/255/255 produces white 255 but still leaves RGB 169/170/171; at 2000 K, an input equal to the computed white basis still leaves RGB 94/50/5. That means sub_ktemp_white cannot actually subtract the configured white cleanly before merge. The fix should scale the white channel as a 0..255 scalar of the chosen white basis, then subtract that same scalar fraction from RGB, instead of normalizing by the sum of the basis channels.
> 
> Maybe you can check, if it is in line with your intend. If the fix does not make sense we can roll it back....
I will take a look into it ASAP.

lougovsk:
@Lord-Grey now regarding what is the right subtraction algorithm. BTH I don't know. I think a simple plausibility test would be:  

1. We have a white of 6,600 Kelvin on our LED strip (255, 255, 253) 
2. We have the same power output on all  LEDs of a RGBW (e.g. 0.3W RGB and 0.1W White) -> full RGB (255,255,255) equals 3x White channel (255)
3. input = (255,255,255)
4. output color ordering = (R,G,B,W)

Now both algos:

1. sumW = 765; fRatio = 1;fWhiteDrive = qBound(0, 1*765, 255) = 255; fActualRatio = 1/3; output = (255 - 255/3, 255 - 255/3, 255 - 255/3 , 255) -> this result kinda makes sense as we expect RGB (255,255,255) to be 3x brighter than W(255)
2. whiteLevel = 1; output = (255 - 255*1,255 - 255*1,255 - 255*1, 255) = (0,0,0,255) -> basically I guess will 3x dimmer than RBG(255,255,255)

So the algo originally implemented was trying to account for brightness difference of 3 RGB LEDs vs  a single white LED. But this also quite a massive assumption, tbh.

Lord-Grey:
How does the output look like in practice with both algorithms?

lougovsk:
I will check out in the evening when it gets dark.

lougovsk:
@Lord-Grey I've done a short round of A/B testing. TBH I didn't much of a difference.

Lord-Grey:
> @Lord-Grey I've done a short round of A/B testing. TBH I didn't much of a difference.

@lougovsk Are you good to go merging the current PR now or would you like to do further testing and/or changes?

PR #2020, opened 2026-06-28
Replace QAction menu with custom QToolButton tray menu

Snaiper-hub:
-


Lord-Grey:
@Snaiper-hub Thanks for providing a PR to further improve and enhance Hyperion's capabilities.
Going forward, please push consistent code which does not require extra compensation actions to get a build done for testing and review.

The PR has following main fixes/improvements included:

1. **Redesigned System Tray Menu** (Windows): Inline icon buttons for Color/Effects/Clear and Suspend/Resume/Restart/Quit. Active source is indicated by a colored strip below the buttons. Dark theme support.
2.  **Persistent Startup Source**: Save the active color or effect as a startup source via the web UI remote control. Hyperion restores it automatically on every start (priority 100), even without opening the web page.
3. **Autorun Toggle Fix**: The Enable/Disable autostart button now correctly reflects the current Windows registry state in real time.
4. **Twilight support on startup**

Given the changes are sometimes interwoven in the commits, let me try to digest and review the individual themes one by one. At a first glance the one or the other might require the design and implementation to be changed.

Lord-Grey:
1. Redesigned System Tray Menu (Windows)

While I appreciate having the layout improved, the main challenge is that the refactor will not work on Linux.
Hyperion runs on multiple platforms and needs to be maintained as such.
Of course there can be platform specific code added, but this will create latter challenges in maintaining the overall code. Therefore, it might be sometimes required to "compromise" to achieve consistency.

On Linux the main entries only show as grey
image

Here is the background why:

Root cause: On GNOME/KDE and most modern Linux desktops, QSystemTrayIcon's context menu isn't drawn by Qt itself — it's exported over D-Bus via the StatusNotifierItem/DBusMenu protocol and rendered natively by the desktop shell. That protocol only understands a plain QAction's text()/icon()/checkable() properties; it has no concept of a custom widget.

The tray menu (introduced in commit a8131e48) built every entry as a QWidgetAction whose icon/text/click-handling lived entirely on a child QToolButton, while the underlying QAction itself was left with blank text and no icon. Under DBusMenu rendering, the child widget is simply never used, so the entries showed up as empty rows that still occupied space — exactly the "three row elements created but empty" you observed (_settingsAction, the first-instance icon row, and the bottom Suspend/Restart/Quit row).

Additionally, the first-instance and bottom rows each squeezed multiple buttons into a single QWidgetAction, which is fundamentally incompatible with DBusMenu since it only supports one icon+label per menu entry — no combination of buttons in one row is possible there.

To maintain compatibility, it would be good to 

a) split the combined Color/Effects/Clear "first instance" row into 3 separate top-level entries (Color action, Effects submenu, Clear action)
b) have a checkmark indicating the "active source"
c) split the combined Suspend/Restart/Quit row is split into 3 separate actions, each with proper icon/text/toggle state.

Something like, the one instance should be on the main level:

image

### Review on Instances

The current code commentary states "First instance: horizontal icon buttons directly in main menu", which is not consistently applied when multiple instances are configured.

Test Sceanrios:

a) Multi instance updates 

Configure 3 instances. Delete the first one.

image

Now add another instance "4th instance" which in the background gets Index 0. Now start the instance.
The Systray is updated ok

image

Stop and start Hyperion

Now the 4th instance is the "first" instance
image

-> From a user's perspective, it is not clear which instance is the one in the main menu in a multi instance setup.

I follow the idea that a user with one instance should not do an extra click, but for many instances the user should also know deterministically which instances an update is applied to.

**Proposal**
I would support a change, where for one instance systems the buttons are on the main level, but for more than one active instance, the systray should then move to row by instance way dynamically.

b) Main Instance sources

The Color/Effect icon displays well, if an Color/Effect executes.
Nevertheless, if no source is available, the Color effect is wrongly highlighted as active (black bar).

image

Lord-Grey:
2. Persistent Startup Source

I assume the background of the feature is that a user can select a running Color or Effect which is then applied at every start-up of Hyperion.

Hyperion already has the capability to define a start-up effect.
So I am wondering why not building up on that feature and make the effect configurable.
Currently, a start-up effect has a duration, but we can discuss to change it to end-less.
This would also addess the complexity introduced in the Hyperion class.
Why should "saved startup source" override "foreground effect if set", if the same problem is to be solved....

Note: The idea of the remote control is to dynamically change configured items.
Every configuration which is applied by standard (e.g. startup source) should always be able to be configured in its own screen too. It is is great to save dynamic setup, but this must be embedded consistently.

Test Scenarios:

a) Current Source is Screen Capture only.

Findings: The "Use Active Source as startup" button is shown to a user and can be pressed.
This results in an update request to the backend.

b) Current Source is an Effect

I find the UI experience misleading with the big button bar below source selection.
I, as a user would expect that I see such a button only next to the sources which can be selected for start-up.

e.g. somewhere here:
image

Lord-Grey:
3. Autorun Toggle Fix

Thank for providing the fix, but I think it introduced a logic bug.
The logic is from a current state view "Autostart Disabled"/"Autostart Enabled" and not from what is the outcome after pressing the button.

```
void SysTray::createBaseTrayMenu()
{
...
	_autorunAction = createAction(tr("&Enable autostart"), ":/autorun.svg", [this]() {
```
That change is fine to have a default consistent with the first startup.
```
	setAutorunState();
	
    void SysTray::setAutorunState()
    {
	    bool currentState = getCurrentAutorunState();
    
            bool SysTray::getCurrentAutorunState()
            {
	            const QSettings reg("HKEY_CURRENT_USER\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Run", QSettings::NativeFormat);
	            bool enabled = (reg.value("Hyperion", 0).toString() == QApplication::applicationFilePath().replace('/', '\\'));
```	            
This results in enabled = true when Hyperion is found (menu action is "Disable Autostart")
, else in false (menu action is "Enable Autostart")            
```	            	            
	            auto* wa = qobject_cast<QWidgetAction*>(_autorunAction);
	            if (wa)
	            {
		            QToolButton* btn = qobject_cast<QToolButton*>(wa->defaultWidget());
		            if (btn)
		            {
			            btn->setText(QStringLiteral("  ") + (!enabled ? tr("&Disable autostart") : tr("&Enable autostart")));
```	
Now you invert the previous logic and create the wrong texts. 
With your logic the text should read (!enabled ? tr("&autostart Disabled") : tr("&autostart Enabled")));
but as we define the actions, the previous code was ok.
```				            
		            }
	            }
	            return enabled;
    
 ```	
"setAutorunState" should be renamed as "toggleAutorunState", then the intend is more obvious.
but result should be ok.
 ```	
	    QSettings reg("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Run", QSettings::NativeFormat);
	    (currentState)
			    ? reg.remove("Hyperion")
			    : reg.setValue("Hyperion", QApplication::applicationFilePath().replace('/', '\\'));
    }

```

Replacing
`bool enabled = (reg.value("Hyperion", 0).toString() == QApplication::applicationFilePath().replace('/', '\\'));`
by 
```
bool enabled = (QString::compare(reg.value("Hyperion").toString(), 
                                 QDir::toNativeSeparators(QApplication::applicationFilePath()), 
                                 Qt::CaseInsensitive) == 0);
```
might further stabalize the call.

libsrc/hyperion/Hyperion.cpp
```diff
@@ -95,10 +110,182 @@ Hyperion::~Hyperion()
 	TRACK_SCOPE_SUBCOMPONENT();
 }
 
+#ifdef _WIN32
+static void enumerateCloudStoreKeys()
```
Lord-Grey, line 114: This is dead code. Please remove.

libsrc/hyperion/Hyperion.cpp
```diff
@@ -95,10 +110,182 @@ Hyperion::~Hyperion()
...
+	return result;
+}
+
 void Hyperion::start()
 {
 	Debug(_log, "Hyperion instance starting...");
 
+#ifdef _WIN32
+	enumerateCloudStoreKeys();
+	{
```
Lord-Grey, line 258: This is dead code. Please remove.

libsrc/hyperion/Hyperion.cpp
```diff
@@ -95,10 +110,182 @@ Hyperion::~Hyperion()
...
+			}
+			RegCloseKey(hSub);
+			subKeySize = 256;
+		}
+		RegCloseKey(hBase);
+	}
+}
+#endif
+
+static QPair<QDateTime, QDateTime> calculateSunriseSunset(double latitude, double longitude, const QDate& date)
```
Lord-Grey, line 183: The Hyperion class has a clear scope. 
Do not bring in other functionality. This should go into a utility class in utils

libsrc/hyperion/Hyperion.cpp
```diff
@@ -148,6 +335,44 @@ void Hyperion::start()
 	connect(this, &Hyperion::compStateChangeRequest, _ledDeviceWrapper.get(), &LedDeviceWrapper::handleComponentState);
 	connect(this, &Hyperion::ledDeviceData, _ledDeviceWrapper.get(), &LedDeviceWrapper::updateLeds);
 
+	// initialize twilight auto-suspend BEFORE device creation to prevent auto-start during daytime
+	{
```
Lord-Grey, line 339: Twighlight handling should go into a separate class, maybe TwilightTransitionManager...

libsrc/hyperion/Hyperion.cpp
```diff
@@ -265,6 +571,17 @@ void Hyperion::handleSettingsUpdate(settings::type type, const QJsonDocument& co
 		updateLedLayout(getSetting(settings::LEDS).array());
 		_ledBuffer.fill(ColorRgb::BLACK, _hwLedCount);
 	}
+	else if (type == settings::TWILIGHT)
+	{
+		QJsonObject const twConfig = config.object();
```
Lord-Grey, line 576: Should be handled by a TwilightTransitionManager

libsrc/hyperion/Hyperion.cpp
```diff
@@ -313,6 +630,11 @@ QJsonDocument Hyperion::getSetting(settings::type type) const
 	return _settingsManager->getSetting(type);
 }
 
+QString Hyperion::getSettingString(settings::type type) const
```
Lord-Grey, line 633: Settings must be always in json-form.
There is not need for a QString getSettingString method.
Please remove.

libsrc/hyperion/SettingsManager.cpp
```diff
@@ -34,6 +34,11 @@ QJsonDocument SettingsManager::getSetting(const QString& type) const
 	return _sTable->getSettingsRecordJson(type);
 }
 
+QString SettingsManager::getSettingString(settings::type type) const
+{
```
Lord-Grey, line 38: Settings must be always in json-form.
There is not need for a QString getSettingString method.
Please remove.

libsrc/hyperion/Hyperion.cpp
```diff
@@ -165,6 +390,82 @@ void Hyperion::start()
 	// initial startup effect
 	hyperion::handleInitialEffect(this, getSetting(settings::FGEFFECT).object());
 
+	// apply saved startup source (overrides foreground effect if set)
+	QJsonDocument startupSrcDoc = getSetting(settings::STARTUPSOURCE);
```
Lord-Grey, line 394: See seperate comment on start-up source

libsrc/hyperion/Hyperion.cpp
```diff
@@ -387,6 +709,77 @@ int Hyperion::isComponentEnabled(hyperion::Components comp) const
 	return _componentRegister->isComponentEnabled(comp);
 }
 
+void Hyperion::checkTwilightState()
```
Lord-Grey, line 712: Twighlight handling should go into a separate class, maybe TwilightTransitionManager...
I would like to understand why you only disable the device and not suspend the whole instance.

TwilightTransitionManager might run as another event service.

assets/webconfig/js/content_remote.js
```diff
@@ -132,6 +132,8 @@ $(document).ready(function () {
     if (prios.length === 0) {
       $('.sstbody').append(`<tr><td colspan="4" class="text-center text-muted">${$.i18n('remote_input_no_sources')}</td></tr>`);
       $('#auto_btn').empty();
+    appendStartupCheckbox();
+    appendTwilightPanel();
```
Lord-Grey, line 136: To be discussed, if the Twilight feature is per instance or a global configuration.

assets/webconfig/js/content_remote.js
```diff
@@ -243,6 +245,46 @@ $(document).ready(function () {
       if ($(this).innerWidth() > maxWidth) maxWidth = $(this).innerWidth();
     });
     $('.btn_input_selection').css("min-width", maxWidth + "px");
+
+    appendStartupCheckbox();
+    appendTwilightPanel();
+  }
+
+  function appendStartupCheckbox() {
```
Lord-Grey, line 253: See the feedback in the separate start-up source comment

include/api/JsonApiCommand.h
```diff
@@ -147,6 +154,7 @@ class SubCommand {
 		case GetSubscriptionCommands: return "getSubscriptionCommands";
 		case GetSubscriptions: return "getSubscriptions";
 		case GetTokenList: return "getTokenList";
+		case GetTwilight: return "get";
```
Lord-Grey, line 157: GetTwilight might be changed to pure "Get"

include/api/JsonApiCommand.h
```diff
@@ -165,6 +173,8 @@ class SubCommand {
 		case Resume: return "resume";
 		case SaveName: return "saveName";
 		case SetConfig: return "setconfig";
+		case SetStartupSource: return "setStartupSource";
+		case SetTwilight: return "set";
```
Lord-Grey, line 177: SetTwilight might be changed to pure "Set"

include/api/JsonApiCommand.h
```diff
@@ -348,6 +358,9 @@ class ApiCommandRegister {
 			{ {"serverinfo", "getSubscriptions"},        { Command::ServerInfo,     SubCommand::GetSubscriptions,        Authorization::Yes,    InstanceCmd::No_or_Single, InstanceCmd::MustRun_Yes,    NoListenerCmd::No  } },
 			{ {"serverinfo", "getSubscriptionCommands"}, { Command::ServerInfo,     SubCommand::GetSubscriptionCommands, Authorization::No,     InstanceCmd::No,           InstanceCmd::MustRun_No,     NoListenerCmd::No  } },
 			{ {"service", "discover"},                   { Command::Service,        SubCommand::Discover,                Authorization::Yes,    InstanceCmd::No,           InstanceCmd::MustRun_No,     NoListenerCmd::Yes } },
+			{ {"startupsource", "get"},                  { Command::StartupSource,  SubCommand::Empty,                   Authorization::Yes,    InstanceCmd::No,           InstanceCmd::MustRun_No,     NoListenerCmd::Yes } },
```
Lord-Grey, line 361: Definitions are incosistent and wrongly defined.
They are InstanceCmd::Yes, there shoudl not be an row with empty sub-command.
the get command has no subcommand defined

Definitions should match the json definition

libsrc/hyperion/Hyperion.cpp
```diff
@@ -768,6 +1161,12 @@ void Hyperion::writeToLeds()
 			}
 		}
 	}
+	else if (_twilightBootComplete && _twilightEnabled && !_isTwilightNight
```
Lord-Grey, line 1164: This requires review after the other fixes were done.
It doies not sound design wise correct.

libsrc/leddevice/LedDeviceWrapper.cpp
```diff
@@ -132,6 +145,14 @@ void LedDeviceWrapper::onIsOnChanged(bool isOn)
 	}
 }
 
+void LedDeviceWrapper::disableDevice()
+{
+	if (_ledDevice)
+	{
+		QMetaObject::invokeMethod(_ledDevice.get(), "disable", Qt::BlockingQueuedConnection);
```
Lord-Grey, line 152: Do not use BlockingQueuedConnection.
We are using Qt's signal and slot mechanisms and LedDeviceWrapper and LedDevice were refactored to do so,
BlockingQueuedConnection shows a design issue.

include/hyperion/SettingsManager.h
```diff
@@ -47,6 +47,11 @@ class SettingsManager : public QObject
 	///
 	QJsonDocument getSetting(settings::type type) const;
 
+	///
+	/// @brief Get raw string value of a setting from the database
+	///
+	QString getSettingString(settings::type type) const;
```
Lord-Grey, line 53: Settings must be always in json-form.
There is not need for a QString getSettingString method.
Please remove.

libsrc/hyperion/Hyperion.cpp
```diff
@@ -165,6 +390,82 @@ void Hyperion::start()
...
+		Info(_log, "Startup source document is null or not an object, skipping");
+	}
+
+	// Apply initial twilight state
+	if (_twilightEnabled)
+	{
+		if (_isTwilightNight)
+		{
+			if (_ledDeviceWrapper)
+				_ledDeviceWrapper->handleComponentState(hyperion::COMP_LEDDEVICE, true);
```
Lord-Grey, line 456: To be reviewed. Likely not fitting in the overall design.
Why is the instance not suspended when twilight is true.

Lord-Grey (changes requested):

Lord-Grey:
4. **Twilight support on startup**

I already provided some comments via the code review above.
In general, the core Twighlight functionality should not be intermingled with the Hyperion scoped code.

But before, exploring the current solution, I would like to understand what user problem you would like to solve?
Should all instances not turn on during daylight or should the feature be applied on selected instances only?

In addition, I suppose that during Daylight the LED device should be off, as well as no extra resources should be consumed.
To achieve the latter, the way is to suspend an (LED-) instance, not to disable an LED-Device.

Let us clarify, what should be achieved and then we can discuss the integration design while the twighlight calculation code etc. can be reused.

Lord-Grey:
@Snaiper-hub In summary, I think you sahred a good amount of code and new features/ideas.
That is great and very much apprechiated!

Given that the current PR contains four independed themes, it might be best if you could issue independed PRs for each change area. Then it is easier to have a ringfenced discussion and review per each.

PR #2030, opened 2026-08-14
Robobloq LedDevice support

immens:
**Summary**

- Support for Robobloq devices

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Build-related changes
- [ ] Other, please describe:

image_2026-08-16_20-08-05

- [ ] Yes
- [x] No

- [x] A convincing reason for adding this feature

**Other information:**

This implementation is based on the reverse-engineering work available in the OpenRGB and hypr-quicklight projects. The code was developed with AI assistance and subsequently reviewed, refined, and adapted to the existing Hyperion architecture.

The implementation has been tested with the 27-inch variant featuring a three-segment strip with 65 LEDs. Other sizes and hardware revisions have not been tested.

For correct LED mapping, the Hyperion LED layout must match the physical arrangement and direction of the strip. The attached screenshot shows the layout used for the tested device.

image_2026-08-14_13-28-16


libsrc/leddevice/dev_hid/ProviderHID.cpp
```diff
@@ -9,6 +9,19 @@
...
 
+namespace
+{
+	QString formatHexValue(unsigned short value)
+	{
+		return QStringLiteral("0x%1").arg(value, 4, 16, QLatin1Char('0'));
+	}
+}
+
+QString ProviderHID::fromWide(const wchar_t* value)
```
Lord-Grey: The method is redundant. QString::fromWCharArray is able to cover nullptr

libsrc/leddevice/dev_hid/ProviderHID.cpp
```diff
@@ -85,9 +98,9 @@ int ProviderHID::open()
 												<< QString("%1").arg(cur_dev->vendor_id, 4, 16, QLatin1Char('0'))
```
Lord-Grey: Should use new formatHexValue method consequently accoss the code

libsrc/leddevice/dev_hid/ProviderHID.cpp
```diff
@@ -85,9 +98,9 @@ int ProviderHID::open()
...
 												<< QString("%1").arg(cur_dev->product_id, 4, 16, QLatin1Char('0'));
 				qCDebug(leddevice_properties()) << "  path:" << cur_dev->path;
```
Lord-Grey: Stablise if path is null

    qCDebug(leddevice_properties()) << "  path:" 
                                    << (cur_dev->path ? cur_dev->path : "NULL");

libsrc/leddevice/dev_hid/ProviderHID.cpp
```diff
@@ -196,33 +209,51 @@ QJsonObject ProviderHID::discover(const QJsonObject& /*params*/)
 {
 	QJsonObject devicesDiscovered;
```
Lord-Grey: Discover should be able to filter via vendorId and/or productId using param

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,566 @@
...
+#include <QJsonArray>
+#include <QStringList>
+#include <QtGlobal>
+
+#include <algorithm>
+#include <limits>
+
+namespace
+{
+	constexpr unsigned short vendorId = 0x1A86;
```
Lord-Grey: Please maintain constants in Uppercase aligning to the coding style

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,566 @@
...
+	constexpr uint8_t initialMessageId = 2;
+
+	constexpr uint8_t actionSetSyncScreen = 0x80;
+	constexpr uint8_t actionReadDeviceInfo = 0x82;
+	constexpr uint8_t actionSetColorRanges = 0x86;
+	constexpr uint8_t actionSetBrightness = 0x87;
+	constexpr uint8_t actionSetOpenUrl = 0x93;
+	constexpr uint8_t actionTurnOffLight = 0x97;
+
+	QString bytesToHex(const uint8_t* data, const int size)
```
Lord-Grey: Suggest you use LedDevice::uint8_t_to_hex_string instead of duplicating code

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,566 @@
...
+{
+	return new LedDeviceRobobloq(deviceConfig);
+}
+
+QJsonObject LedDeviceRobobloq::discover(const QJsonObject& /*params*/)
+{
+	QJsonObject result;
+	result.insert("ledDeviceType", _activeDeviceType);
+
+	QJsonArray deviceList = enumerateHidDevices(vendorId, productId, usagePage, usage);
```
Lord-Grey: Discover should not get device details yet.

The logic is
1. discover to get a list of relevant devices
The UI then triggers 2. or 3. with a selected device
2. getProperties for a given device
3. optional, identify a given device

You might want to call discover of the parent class with providing vendor and productID via params (see my other comment)

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,566 @@
...
+}
+
+QJsonObject LedDeviceRobobloq::getProperties(const QJsonObject& params)
+{
+	QJsonObject result;
+	QJsonObject properties;
+
+	if (const QString devicePath = params["output"].toString(); !devicePath.isEmpty())
+	{
+		const QJsonArray devices = enumerateHidDevices(vendorId, productId, usagePage, usage);
```
Lord-Grey: Why do you enumerate devices again?
Cann the device proverties not accessed by the given device path?

libsrc/leddevice/dev_hid/LedDeviceRobobloq.h
```diff
@@ -0,0 +1,62 @@
...
+	bool powerOff() override;
+
+private:
+	struct DeviceInfo
+	{
+		QString _deviceId;
+		QString _uuid;
+		QString _firmwareVersion;
+		uint8_t _physicalSize{0};
+		uint8_t _ledCount{0};
```
Lord-Grey: Suggest renaming it to _hwLedCount, as _ledCount from LedDevice is hidden

libsrc/leddevice/dev_hid/LedDeviceRobobloq.h
```diff
@@ -0,0 +1,62 @@
...
+protected:
+	bool init(const QJsonObject& deviceConfig) override;
+	int open() override;
+	int write(const QVector<ColorRgb>& ledValues) override;
+	bool powerOff() override;
+
+private:
+	struct DeviceInfo
+	{
+		QString _deviceId;
```
Lord-Grey: Why are elements of DeviceInfo name with "_"? They are not members, but the struct will be

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+}
+
+int LedDeviceRobobloq::open()
+{
+	ProviderHID::close();
+
+	Info(_log, "Opening Robobloq HID device at path '%s'", QSTRING_CSTR(_devicePath));
+
+	// Locate the protocol HID collection matching the path selected during discovery
+	hid_device_info* devices = hid_enumerate(vendorId, productId);
```
Lord-Grey: why is the ProviderHID::open() not good enough. It looks like too much code duplication and ignoring polimorphism.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+	_deviceHandle = hid_open_path(selectedPath.constData());
+	if (_deviceHandle == nullptr)
+	{
+		setInError(
+			QStringLiteral("Failed to open Robobloq HID path '%1'. Check device permissions.").arg(_devicePath));
+		return -1;
+	}
+
+	_isDeviceReady = true;
```
Lord-Grey: I would assume that the device will not work, if the subsequent code getting deviceInfo is not successful.
Therefore setting _isDeviceReady = true is too early.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+		Warning(_log, "Failed to disable the Robobloq URL shortcut");
+		if (_deviceHandle == nullptr)
+		{
+			setInError(QStringLiteral("Robobloq HID connection was lost while disabling the URL shortcut"));
+			return -1;
+		}
+	}
+
+	// Apply the configured hardware settings before accepting LED updates
+	if (sendRb(actionSetBrightness, {static_cast<uint8_t>(_hardwareBrightness)}) < 0)
```
Lord-Grey: Some devices turn on when setting the brightness. Therefore, it is advisable to set this in the powerOn method.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+			static_cast<int>(ledValues.size()), maxLedCount);
+		return -1;
+	}
+
+	return sendSc(actionSetSyncScreen, encodeColors(ledValues));
+}
+
+bool LedDeviceRobobloq::powerOff()
+{
+	if (_isStayOnAfterStreaming)
```
Lord-Grey: This is dead code. There is nothing the json config schema that allows configuring it

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	{
+		return true;
+	}
+
+	const bool effectStopped = sendRb(actionTurnOffLight) == 0;
```
Lord-Grey: See LedDevice::powerOff(), there will be black LED written. 
You might need to do

trackDevice(leddevice_flow, "Power OFF") << ", is device ready :" << (_isDeviceReady ? "YES" : "NO");
if (!_isDeviceReady)
{
    return false;
}

writeBlack();
return (sendRb(actionTurnOffLight) == 0);

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+			if (received < 0)
+			{
+				return false;
+			}
+			if (received == 0)
+			{
+				break;
+			}
+
+			Debug(_log, "Robobloq device-info reply (%d bytes): %s",
```
Lord-Grey: To trace low level protocol requests and responses, I suggest you use
qCDebug(leddevice_properties) << QString("#[0x%1], Data returned: [%2]").arg(QString::number(_sequenceNumber, 16), response.toHex());

See LedDeviceCololight for illustration or other LedDevices

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	// SC: magic, big-endian uint16 length, message ID, action, payload, checksum.
+	message[0] = 'S';
+	message[1] = 'C';
+	message[2] = static_cast<uint8_t>((logicalLength >> 8) & 0xFF);
+	message[3] = static_cast<uint8_t>(logicalLength & 0xFF);
+	message[4] = allocateMessageId(_nextMessageId);
+	message[5] = action;
+	std::copy(payload.cbegin(), payload.cend(), message.begin() + 6);
+	message[logicalLength - 1] = checksum(message.constData(), logicalLength - 1);
+	const QVector<uint8_t> reports = padToHidReports(message);
```
Lord-Grey: maybe the function can be optimized. it looks like mulitiple iterations and appending when also looking at the sub functions called.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+bool LedDeviceRobobloq::readDeviceInfo(hid_device* handle, uint8_t& nextMessageId, DeviceInfo& deviceInfo) const
+{
+	if (handle == nullptr)
+	{
+		return false;
+	}
+
+	uint8_t staleReport[hidReportSize + 1] = {};
+	while (hid_read_timeout(handle, staleReport, sizeof(staleReport), 0) > 0)
+	{
```
Lord-Grey, line 360: add a comment why the statement block is empty

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	}
+
+	const QVector<uint8_t> message = padToHidReports(
+		createRbMessage(action, payload, allocateMessageId(_nextMessageId)));
+	return writeBytes(static_cast<unsigned>(message.size()), message.constData()) < 0 ? -1 : 0;
+}
+
+int LedDeviceRobobloq::sendSc(uint8_t action, const QVector<uint8_t>& payload)
+{
+	const int logicalLength = 7 + payload.size();
```
Lord-Grey: change int to auto

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+QJsonObject LedDeviceRobobloq::getProperties(const QJsonObject& params)
+{
+	QJsonObject result;
+	QJsonObject properties;
+
+	if (const QString devicePath = params["output"].toString(); !devicePath.isEmpty())
+	{
+		const QJsonArray devices = enumerateHidDevices(vendorId, productId, usagePage, usage);
+		for (const auto& device : devices)
```
Lord-Grey: getProperties should be for one device

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
```diff
@@ -0,0 +1,569 @@
...
+	QJsonObject hidConfig(deviceConfig);
+	hidConfig["VID"] = QStringLiteral("0x%1").arg(vendorId, 4, 16, QLatin1Char('0'));
+	hidConfig["PID"] = QStringLiteral("0x%1").arg(productId, 4, 16, QLatin1Char('0'));
+
+	if (!ProviderHID::init(hidConfig))
+	{
+		return false;
+	}
+
+	_devicePath = deviceConfig["output"].toString();
```
Lord-Grey: why path? Other HID devices work via VID and PID...
Is the problem that there might be multiple devices?
If yes, why is then the ProviderHID not extended to make the functionality generally available.
LedDeviceRobobloq should only have device specific implementations not HID basics

Lord-Grey:
@immens Thank you for your contribution making Robobloq devices available to the community.
I did a code review and shared some suggestions/questions via the review comments above.

In case you have questions or need clarification, do not hesitate to reach out.

immens:
> @immens Thank you for your contribution making Robobloq devices available to the community. I did a code review and shared some suggestions/questions via the review comments above.
> 
> In case you have questions or need clarification, do not hesitate to reach out.

Unfortunately, I can't see any comments other than the one from the bot. I addressed that one by adding three comments to the code and squashed the commits.

Could the review comments still be pending?

Lord-Grey:
> Could the review comments still be pending?

If I look directly in the PR on GitHub, I already see the comments below the on from the bot.
If you working via an IDE, you might need to refresh the PR?

immens:
> > Could the review comments still be pending?
> 
> If I look directly in the PR on GitHub, I already see the comments below the on from the bot. If you working via an IDE, you might need to refresh the PR?

I'm viewing the PR in the browser, but unfortunately I still can't see any review comments. Could I have broken something by force-pushing after squashing the commits?

Lord-Grey (changes requested):

Lord-Grey:
> > > Could the review comments still be pending?
> > 
> > 
> > If I look directly in the PR on GitHub, I already see the comments below the on from the bot. If you working via an IDE, you might need to refresh the PR?
> 
> I'm viewing the PR in the browser, but unfortunately I still can't see any review comments. Could I have broken something by force-pushing after squashing the commits?

My bad! Looks like I missed to submit the review.....
Can you see the feedback now?

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
immens: `LedDevice::uint8_t_to_hex_string()` inserts separators between bytes, while the device ID and UUID are displayed as compact identifiers

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
immens: The `0x82` query during discovery is intentional. The tested device exposes generic HID metadata, which may not uniquely identify multiple units (e.g. `S/N 0123456789`). The protocol response provides the UUID/device ID and hardware characteristics used for a unique and informative selection label

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
immens: The base implementation is not equivalent here: `writeBlack()` dispatches to the virtual Robobloq `write()` and produces an `SC 0x80` frame for the configured LEDs. The reverse-engineered shutdown sequence is `RB 0x97` followed by `RB 0x86` color ranges, including a trailing black range through `0xFE`

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
immens: Yes, both cases apply. The device exposes at least two HID collections under the same VID/PID — the keyboard collection (`0001:0006`) and the protocol collection (`ff00:0001`) — and multiple physical units may also be connected. `hid_open(VID, PID, nullptr)` therefore cannot select the target unambiguously

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
Lord-Grey: I would suggest, we update LedDevice::uint8_t_to_hex_string() to have no separators by default.
A separator was only used optional for Debugging purposes.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
Lord-Grey: ProviderHID::enumerateHidDevices already has S/N etc.
 My feedback is related that LedDeviceRobobloq::buildDeviceProperties invokes readDeviceInfo which is requesting RoboBloq device specifics.
 Since I do not know what a user knows about a Robobloq, maybe DeviceID is a topic if it identifies a type? (here a type name would be better then a number).
 But I doubt that a user would select a device by its UUID or firmware version.

 Background of my question is that a discover call has a timeout period and if you put too many requests for detailed inforamtion for all devices the discovery might fail. My feedback is was just food for thought... (an to keep focus on what a user "really" require to identify the device).

 Normally, when the device list ist populated, the first entry triggers the getProperties for the given device and updates the screen.

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
Lord-Grey: Yes, that is what I assumed too.
Nevertheless the _devicePath and teh handling should go into ProviderHID as it is relevant for all HID devices...

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
Lord-Grey: Thanks for the explainaition.

You still might want to add

trackDevice(leddevice_flow, "Power OFF") << ", is device ready :" << (_isDeviceReady ? "YES" : "NO");
if (!_isDeviceReady)
{
return false;
}

to saveguard the sendRb command....

immens:
I’ve addressed some of the review comments. After digging a bit deeper into the project, I also did a small refactor of `ProviderHID` to reuse its discovery logic here.

It looks like the default VID/PID values probably don’t belong in a generic provider, but I’ve left them unchanged for now since this needs clarification. @Lord-Grey, could you please take a look?

Lord-Grey:
> I’ve addressed some of the review comments. After digging a bit deeper into the project, I also did a small refactor of `ProviderHID` to reuse its discovery logic here.

Thank you for addressing various comments! At a first glance things looks good, but I would like to dig into the updated code. Unfortunately, I will be traveling soon. Therefore, please already find my apologies that it might require some time...

> It looks like the default VID/PID values probably don’t belong in a generic provider, but I’ve left them unchanged for now since this needs clarification. @Lord-Grey, could you please take a look?

Yes. I stumbled of the defaults too. I agree that they look like values "to have values" and do not make sense....
Might be better to have them defaulted to 0x0000 to document nothing has been set.

```
Vendor ID 0x2341 → Arduino SA / Arduino
Product ID 0x8036 → Arduino Leonardo
```

libsrc/leddevice/dev_hid/LedDeviceRobobloq.cpp
immens: I can’t say I love it — the HID metadata shown in the dropdown isn’t particularly useful for distinguishing between devices — but it’s probably acceptable overall

<It wouldn’t let me attach the screenshot here, so I updated the comment directly in the PR instead>

Lord-Grey: Have you had a look at other devices, which have an identify button?

I think this would be more userfriendly than lots of identifiers...
You could have unique device identifiers in the list and then the user press identify and the device blinks.
If it is the correct one, user saves.

Look at the screen from a user perspective, not from an engineer 😄 

Note: Try the UI screen on a mobile too. The relevant select item must fit there too (and should be readable)

immens: That’s actually what I’m trying to do 🙂 I had another look through the existing device implementations. Device properties are generally used to populate the relevant configuration fields rather than being displayed as a metadata dump.

For the remaining useful hardware information, the existing Information panel below the settings seems like the most appropriate place. I’m thinking of keeping only a short summary there, such as the detected screen diagonal and LED count. Would that be acceptable?

immens: Ended up adjusting the layout a little. This is what it looks like on mobile now:

IMG000 (2)

immens:
Hello @Lord-Grey, are there any other changes I should make?

Lord-Grey:
@immens I am sorry not coming back to you since a while; too many other things came into my way outside the Hyperion project.
Let me find some time in the course of the week to have another look at your PR.

Lord-Grey:
@immens I had another round of review (and testing without a Robobloq device).
I did some small code updates myself. Hoping that is OK with you.

Could you do another round of testing yourself with an existing device?
If that works, you are ok with the code changes, I am happy to merge the PR.

### **Changes done** 
- HID usage and usage page is not supported on Linux (see https://github.com/libusb/hidapi/blob/852cc68b8e0e4c7eba0815ef992951cd8b75c55b/hidapi/hidapi.h#L173)
#### UI
- Added a dynamic Robobloq device tile string in the UI
- Disabled HW-LED count, if resolved from device to prevent users doing a wrong config
- Generalized code on updated info_container_text, if no item is selected
- Removed LED count in info text, as the dedicated HW LED count was updated already
(Going forward we might want to have a dedicated device properties element which every device can update dynamically, but for now it is a good suggestion which stipulates the same for other devices)
 

**Additional feedback:**
You might want to review the IDENTIFY_INTERVAL{500}. Currently, the identify roundtrip would be minimum 2s given the two blinks. That might result that the identify call from the browser will time-out.
Maybe you could check, if there would be any issues in the UI during device identification.

Lord-Grey (approved):

immens:
> @immens I had another round of review (and testing without a Robobloq device). I did some small code updates myself. Hoping that is OK with you.
> 
> Could you do another round of testing yourself with an existing device? If that works, you are ok with the code changes, I am happy to merge the PR.
> ### **Changes done**
> 
>     * HID usage and usage page is not supported on Linux (see https://github.com/libusb/hidapi/blob/852cc68b8e0e4c7eba0815ef992951cd8b75c55b/hidapi/hidapi.h#L173)
> 
> 
> #### UI
> 
>     * Added a dynamic Robobloq device tile string in the UI
> 
>     * Disabled HW-LED count, if resolved from device to prevent users doing a wrong config
> 
>     * Generalized code on updated info_container_text, if no item is selected
> 
>     * Removed LED count in info text, as the dedicated HW LED count was updated already
>       (Going forward we might want to have a dedicated device properties element which every device can update dynamically, but for now it is a good suggestion which stipulates the same for other devices)
> 
> 
> **Additional feedback:** You might want to review the IDENTIFY_INTERVAL{500}. Currently, the identify roundtrip would be minimum 2s given the two blinks. That might result that the identify call from the browser will time-out. Maybe you could check, if there would be any issues in the UI during device identification.



There is definitely a problem due to the removal of usage and the usage page.

Identification seems to work fine (at least while the device isn’t involved in output). I’ll test it more later.

Lord-Grey:
@immens Have you had a chance looking into the issue that we can put the PR to bed?
I would assume that the duplicate entry is not related to the removal of usage and the usage page, but could be a UI glich...
Suggest you check what is send to the UI during discovery. Is that one or two records...?

immens:
> @immens Have you had a chance looking into the issue that we can put the PR to bed? I would assume that the duplicate entry is not related to the removal of usage and the usage page, but could be a UI glich... Suggest you check what is send to the UI during discovery. Is that one or two records...?

I tested it for a while, and it works just as it did before the changes. I haven't noticed any issues so far.

I also checked the discovery path: two records are sent to the UI, not one. The device exposes two HID interfaces with the same VID/PID. Previously, they were filtered by `usagePage` and `usage`, so only the `0xFF00:0x0001` interface was returned.

In the current code, this filtering is guarded by `#if defined(MACOS) || defined(WINDOWS)`. I couldn't find either macro being defined in the project — Qt uses `Q_OS_MACOS` and `Q_OS_WIN` — so the filtering is also skipped in my Windows build. As a result, both HID interfaces are sent to the UI and displayed as separate entries. Therefore, this does not appear to be a UI glitch.

immens:
@Lord-Grey, just a heads-up: the latest commit adds `interfaceNumber` filtering for Linux and restores usage-based filtering on Windows and macOS using the Qt platform macros. Discovery now returns a single device entry on Windows, and everything works as expected. Unfortunately, I don't have a way to test it on Linux.

Lord-Grey:
> @Lord-Grey, just a heads-up: the latest commit adds `interfaceNumber` filtering for Linux and restores usage-based filtering on Windows and macOS using the Qt platform macros. Discovery now returns a single device entry on Windows, and everything works as expected. Unfortunately, I don't have a way to test it on Linux.

Thanks for fixing the wrong marco definitions.
I just had to use raw compiler macros, not the Q_OS_ ones, given I had challanges on the Linux IDE and it would be consistent with the rest of the project's code.

PR #2033, opened 2026-08-16
Add gamescope and xdg-desktop-portal screen grabbers for Wayland

### LINKED ISSUE #1096: X86_64 Linux Wayland Session - no Screen grabbing.
- [x] I confirm that this is an issue rather than a question.

## Bug report
When i start a Wayland Screen Session under Manjaro with Gnome Desktop, the Screen grabbing does not work, no matter which grabber I use.

#### Steps to reproduce
Instead of X.org use Wayland for Screen session.

#### What is expected?
That the function is implemented under Wayland, Wayland Support for Hyperion.

#### What is actually happening?
No Screen grabbing !

#### System
Hyperion 2.0.0-alpha.8
Manjaro 20.1.2 testet with Kernel 5.8 and 5.9
Graphic AMD Ryzen 3 3200G

bhill00:
# Add gamescope and xdg-desktop-portal screen grabbers for Wayland

Fixes #1096

## Problem

Hyperion.NG has no working screen-capture path on Wayland today, no matter which grabber you pick (exactly what #1096 describes):

- `X11Grabber`/`XcbGrabber`/`QtGrabber` explicitly check for Wayland and refuse to run.
- `DRMFrameGrabber` needs to be the sole DRM master, which it can't be while a compositor (KWin, GNOME Shell, etc.) is running.

This adds two new grabbers that cover the two situations that actually come up in practice.

## What's added

**`GamescopeGrabber`** (config value `"gamescope"`) — captures gamescope's composited output directly. Gamescope sets `XDG_SESSION_TYPE=x11` for legacy X11 game compatibility even though it's a Wayland compositor, and it doesn't run an `xdg-desktop-portal` ScreenCast backend — but it does expose its output as a plain PipeWire node literally named `"gamescope"` (the same mechanism `obs-vkcapture` uses for Steam Deck Game Mode capture). No portal permission dialog, no D-Bus involved.

**`DesktopPortalGrabber`** (config value `"desktop-portal"`) — captures a normal desktop session (KDE Plasma, GNOME, ...) via `xdg-desktop-portal`'s `ScreenCast` interface, the standard permission-gated mechanism every other Wayland screen-capture consumer uses. Negotiates over D-Bus (`CreateSession` → `SelectSources` → `Start` → `OpenPipeWireRemote`) using Qt's own `QtDBus` module — already an optional dependency of this project (`src/hyperiond/CMakeLists.txt` already conditionally links it for the events subsystem), so no new `libportal` dependency. The one-time permission grant is persisted via a `restore_token` so it survives daemon restarts without re-prompting.

Both grabbers:
- Run a persistent background thread with a discover → stream → retry loop rather than a one-shot check at construction, so `hyperiond` can run continuously and pick up sessions appearing/disappearing (a gamescope game starting/stopping, portal permission granted/revoked) without an external watcher or restart.
- Negotiate the mandatory `LINEAR` DRM modifier and prefer 4-byte-aligned packed RGB formats (`BGRx`/`RGBx`/`BGRA`), and handle `SPA_DATA_DmaBuf` buffers (not just `SPA_DATA_MemFd`) with a manual mmap fallback — needed for real GPU-rendered game/desktop content, not just synthetic test apps like `glxgears`.
- Are wired into the existing grabber-selection plumbing the same way every other grabber is — `GrabberWrapper::availableGrabbers()`, the `updateScreenGrabbers()`/`evalScreenGrabberType()` dispatch in `hyperiond.cpp`, and `JsonInfo::discoverScreenInputs()` (a separate registration that specifically feeds the web UI's "Device discovered" dropdown — easy to miss if you only wire up `availableGrabbers()`).

Gated behind `ENABLE_GAMESCOPE` and `ENABLE_DESKTOP_PORTAL` CMake options (both default `ON` on Linux), each pulling in `libpipewire-0.3` via pkg-config; `DesktopPortalGrabber` also needs `Qt::DBus`.

## Usage

**Gamescope**: launch your game under gamescope, either via Steam's built-in "Force use of Gamescope" launch option, or manually with e.g. `gamescope -W <width> -H <height> -f -- %command%` in the game's launch options. Then in Hyperion's web UI, set the screen grabber to **Gamescope** (Configuration → Capturing Hardware → Screen Capture). No other setup needed — the grabber auto-detects the gamescope session whenever one exists and goes back to waiting when it ends.

**Desktop Portal**: set the screen grabber to **Desktop Portal**. The first time it activates, your desktop's portal implementation will show a one-time screen-share permission dialog — approve it, and it won't ask again (the grant is remembered via a restore token).

## How to test this PR

1. On a Wayland session (KDE Plasma or GNOME), build with `-DENABLE_GAMESCOPE=ON -DENABLE_DESKTOP_PORTAL=ON` (default on Linux).
2. For gamescope: launch anything under gamescope (`gamescope -- glxgears` is enough to prove it works), set the grabber to `gamescope`, confirm LEDs/output respond.
3. For desktop-portal: set the grabber to `desktop-portal`, approve the permission dialog, confirm output responds to on-screen content, then restart `hyperiond` and confirm it reconnects without a second prompt.

## Update (2026-08-21): also verified on NVIDIA

All prior testing here was on AMD/Mesa (Framework desktop, Strix Halo APU). Moving the same OS
drive to an NVIDIA RTX 3090 (proprietary driver) surfaced a real gap: `desktop-portal`'s format
request demanded a mandatory `LINEAR` DRM modifier for its DMA-BUF request. AMD/Mesa happily
grants that; NVIDIA's driver does not — KWin's screencast source there only ever offers its own
tiled/compressed modifiers, never `LINEAR`. Negotiation failed every single time
(`no more input formats`, confirmed via the PipeWire debug log), and the grabber's retry loop
(every 2s, indefinitely) both pinned real CPU and — on KDE Plasma — repeatedly flashed the
screen-share tray indicator on and off as each attempt opened and immediately tore down a
ScreenCast session.

Latest commit adds a second, lower-priority format alternative with no modifier property at all,
which tells PipeWire this client will also accept a non-DMA-BUF (plain memory) buffer. Confirmed
via `pw-dump` on the RTX 3090: negotiation now succeeds through that fallback — the resulting
Format has no `modifier` key (plain `BGRx`, native resolution), and the PipeWire link shows
`active` rather than retrying. So KWin's NVIDIA-backed screencast implementation does support
handing back a non-tiled buffer — it just won't offer one unless a client explicitly signals it's
willing to accept it. `onStreamProcess()` already branched on `SPA_DATA_MemFd`/`SPA_DATA_DmaBuf`
for the buffer-mapping path, so no change was needed there.

Not yet tested: Intel GPUs, or NVIDIA under GNOME's portal backend rather than KDE's.

## Known limitations (being upfront about these)

- `desktop-portal` reads DMA-BUF/GPU memory from the CPU on AMD/Mesa, which is meaningfully slower than a normal RAM copy on that hardware (write-combined/uncached mapping). Mitigated with a consumer-side frame-rate throttle and a sequential-read strategy that keeps CPU cost bounded, but it's still non-trivial at higher configured capture rates — see the PR's commit history / linked writeup for the full investigation if this matters for your use case. (On NVIDIA, via the non-DMA-BUF fallback above, this specific cost doesn't apply the same way — not yet characterized separately.)
- Neither grabber has a graceful mid-negotiation cancellation path if `hyperiond` is asked to shut down while a `desktop-portal` permission dialog is still unanswered — this can make the daemon slow to stop in that specific window. Not expected to matter in normal use (the dialog only appears once, ever, thanks to the restore token) but worth knowing about.
- Not tested on GNOME's own portal backend (`xdg-desktop-portal-gnome`) — only KDE's (`xdg-desktop-portal-kde`). The D-Bus interface is the same standardized one, so it should work, but I haven't verified it myself.


CMakeLists.txt
```diff
@@ -716,6 +750,14 @@ find_package(Qt${QT_VERSION_MAJOR} ${QT_VERSION} COMPONENTS Core Gui Network Sql
 
 message(STATUS "Qt version used: ${QT_VERSION}")
 
+if(ENABLE_DESKTOP_PORTAL)
+	find_package(Qt${QT_VERSION_MAJOR} COMPONENTS DBus QUIET)
```
Lord-Grey, line 754: I think the DBus dependency should go into the grabber related CMakeList files.

libsrc/api/CMakeLists.txt
```diff
@@ -23,6 +23,26 @@ if(ENABLE_DRM)
 	target_link_libraries(hyperion-api drm-grabber)
 endif(ENABLE_DRM)
 
+if(ENABLE_GAMESCOPE)
+	# GrabberConfig.h (included by JsonInfo.cpp for discoverGrabber<GamescopeGrabber>())
```
Lord-Grey, line 27: The dependency should go into the grabber related CMakeList files.
And include_directories and target_link_libraries should be similar as above

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,882 @@
...
+				SPA_VIDEO_FORMAT_BGRx,
+				SPA_VIDEO_FORMAT_RGBx,
+				SPA_VIDEO_FORMAT_BGRA
+			),
+			SPA_FORMAT_VIDEO_size, SPA_POD_CHOICE_RANGE_Rectangle(&rectDefault, &rectMin, &rectMax),
+			SPA_FORMAT_VIDEO_framerate, SPA_POD_CHOICE_RANGE_Fraction(&rateDefault, &rateMin, &rateMax),
+			// Same LINEAR-modifier requirement as GamescopeGrabber - a normal compositor's
+			// composited output can be a tiled/compressed DMA-BUF just as easily as
+			// gamescope's, and this code has no way to decode that layout.
+			SPA_POD_Propf(SPA_FORMAT_VIDEO_modifier, SPA_POD_PROP_FLAG_MANDATORY, SPA_POD_CHOICE_ENUM_Long(2, 0L, 0L))
```
Lord-Grey: When building under Ubuntu 24.04, I get the following compile error:
‘SPA_POD_Propf’ was not declared in this scope; did you mean ‘SPA_POD_Prop’?

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+			// Real GPU-rendered content (an actual game, as opposed to a simple
+			// test app) gets composited into GPU-native DMA-BUF buffers, which
+			// may use a tiled/compressed modifier this code has no way to
+			// decode. Explicitly requiring the LINEAR modifier (value 0 - see
+			// DRM_FORMAT_MOD_LINEAR in drm_fourcc.h, not included here to avoid
+			// a new libdrm dependency for one well-known constant) makes
+			// gamescope either hand back a plain CPU-readable buffer or fail to
+			// negotiate a format cleanly, instead of negotiating successfully
+			// and then failing later at the buffer-import step.
+			SPA_POD_Propf(SPA_FORMAT_VIDEO_modifier, SPA_POD_PROP_FLAG_MANDATORY, SPA_POD_CHOICE_ENUM_Long(2, 0L, 0L))
```
Lord-Grey: When building under Ubuntu 24.04, I get the following compile error:

‘SPA_POD_Propf’ was not declared in this scope; did you mean ‘SPA_POD_Prop’?

Changing to suggested ‘SPA_POD_Prop’ let the build succeed.

src/hyperiond/hyperiond.cpp
```diff
@@ -866,11 +878,49 @@ QString HyperionDaemon::evalScreenGrabberType()
 {
```
Lord-Grey, line 878: Suggest to treamlien the function as 
```suggestion
QString HyperionDaemon::evalScreenGrabberType()
{
#ifdef ENABLE_GAMESCOPE
    // gamescope -> GAMESCOPE_WAYLAND_DISPLAY is set inside gamescope sessions
    if (!qEnvironmentVariableIsEmpty("GAMESCOPE_WAYLAND_DISPLAY"))
    {
        return "gamescope";
    }
#endif

#ifdef ENABLE_DESKTOP_PORTAL
    // desktop-portal -> real Wayland session
    if (!qEnvironmentVariableIsEmpty("WAYLAND_DISPLAY"))
    {
        return "desktop-portal";
    }
#endif

    // dispmanx -> Raspberry Pi
    if (QFile::exists("/dev/vchiq"))
    {
        return "dispmanx";
    }

    // amlogic -> Amlogic devices
    if (QFile::exists("/dev/amvideo"))
    {
        static const QString amlDevice("/dev/amvideocap0");
        if (!QFile::exists(amlDevice))
        {
            Error(_log, "grabber device '%s' for type amlogic not found!", QSTRING_CSTR(amlDevice));
        }
        return "amlogic";
    }

    // X11 / XCB fallback if DISPLAY is set
    if (!qEnvironmentVariableIsEmpty("DISPLAY"))
    {
#if defined(ENABLE_X11)
        return "x11";
#elif defined(ENABLE_XCB)
        return "xcb";
#endif
    }

    // Default fallback
    return "qt";
}
```

CMakeLists.txt
```diff
@@ -139,6 +139,8 @@ set(DEFAULT_V4L2                        OFF)
 set(DEFAULT_AUDIO                       ON)
 set(DEFAULT_X11                         OFF)
 set(DEFAULT_XCB                         OFF)
+set(DEFAULT_GAMESCOPE                    OFF)
```
Lord-Grey: Please make indents consistent accross the file

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+	constexpr auto RETRY_INTERVAL = std::chrono::milliseconds(2000);
+}
+
+GamescopeGrabber::GamescopeGrabber(int cropLeft, int cropRight, int cropTop, int cropBottom)
+	: Grabber("GAMESCOPE", cropLeft, cropRight, cropTop, cropBottom)
+{
+	// Always available: whether a gamescope session is currently running is a
+	// transient condition the background thread keeps retrying against, not a
+	// platform capability decided once at startup.
+	_isAvailable = true;
```
Lord-Grey, line 24: Suggest you add some tracing, which can be enabled via qtlogging.ini in case further insights into the grabber's activity flow is requried. see QtGrabber as reference.

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+	_thread.emplace(&GamescopeGrabber::pipewireThreadMain, this);
+}
+
+GamescopeGrabber::~GamescopeGrabber()
+{
+	stop();
```
Lord-Grey, line 31: Likly that this will not work, as the object is destroyed later via QT.

```suggestion
	TRACK_SCOPE();
    stop();
```

to check when and if stop() is called.
To see the object lifecycle, enable 

# Object tracing
hyperion.objects.track.debug=true

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+void GamescopeGrabber::onStreamStateChanged(void* userdata, enum pw_stream_state /*old*/, enum pw_stream_state state, const char* error)
+{
+	auto* self = static_cast<GamescopeGrabber*>(userdata);
+
+	if (state == PW_STREAM_STATE_ERROR || state == PW_STREAM_STATE_UNCONNECTED)
+	{
+		if (error != nullptr)
+		{
+			Warning(self->_log, "[pipewire] gamescope stream ended: %s", error);
```
Lord-Grey: Is this more an Info rather Warning?

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+}
+
+void GamescopeGrabber::pipewireThreadMain()
+{
+	pw_init(nullptr, nullptr);
+
+	if (!connectCore())
+	{
+		pw_deinit();
+		return;
```
Lord-Grey, line 481: Should the Grabber not be setInError()? Or is the error recoverable by the grabber itself?

libsrc/grabber/gamescope/GamescopeGrabber.cpp
```diff
@@ -0,0 +1,494 @@
...
+	defaults["video_input"] = videoInputDefault;
+	inputsDiscovered["default"] = defaults;
+
+	return inputsDiscovered;
+}
+
+void GamescopeGrabber::onCoreError(void* userdata, uint32_t id, int seq, int res, const char* message)
+{
+	auto* self = static_cast<GamescopeGrabber*>(userdata);
+	Error(self->_log, "[pipewire] Error id: %d seq: %d res: %d (%s) %s", id, seq, res, strerror(res), message);
```
Lord-Grey, line 128: Are those Errors recoverable by the grabber? If not do a setInError.

And do a "if (_isDeviceInError)" where applicable...

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,882 @@
...
+}
+
+int DesktopPortalGrabber::grabFrame(Image<ColorRgb>& image, bool /*forceUpdate*/)
+{
+	if (!_connected.load())
+	{
+		return -1;
+	}
+
+	std::lock_guard<std::mutex> lock(_bufferMutex);
```
Lord-Grey: Why are not Qt's mutex capabilities used?

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,882 @@
...
+	QJsonObject defaults;
+	defaults["video_input"] = videoInputDefault;
+	inputsDiscovered["default"] = defaults;
+
+	return inputsDiscovered;
+}
+
+QString DesktopPortalGrabber::restoreTokenPath() const
+{
+	return QDir::homePath() + QStringLiteral("/.hyperion/desktop-portal-restore-token");
```
Lord-Grey: Need to revist, as Hyperion's configuration/runtime directory cann not be assumed to be fix $HOME/.hyperion.
Better, store it in AppDataLocation (see ProviderRestApi::matchesPinnedCertificate)


```suggestion
	QString const appDataDir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
	QString const tokenDir = appDataDir + "/token";
```

Best would be having the token to be stored within the configuration db where the setup is stored?

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,882 @@
...
+	QSize size = getScreenSize();
+	if (size.width() == 0 || size.height() == 0)
+	{
+		size = QSize(1920, 1080);
+	}
+
+	QJsonObject resolution;
+	resolution["width"] = size.width();
+	resolution["height"] = size.height();
+	resolution["fps"] = getFpsSupported();
```
Lord-Grey, line 228: If the fps rate is "Capped at 30fps" as in the commentary below, the fpsSupported needs to be set to values other than the default rates.

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,882 @@
...
+		pw_stream_queue_buffer(self->_stream, pwBuffer);
+		return;
+	}
+	lastProcessed = throttleNow;
+
+	spa_buffer* spaBuffer = pwBuffer->buffer;
+
+	const auto width = static_cast<int>(self->_format.info.raw.size.width);
+	const auto height = static_cast<int>(self->_format.info.raw.size.height);
+	const auto spaFormat = self->_format.info.raw.format;
```
Lord-Grey, line 575: Could the frames not already be requested as RGB32 during capturing that the transforamtion is done by the GPU?
if _imageResampler.processImage could be avoided, woudl be best (e.g. see DDA Grabber)

Lord-Grey (commented):
Please find some feedback/questions as per an initial review. Some comments are applicable for both grabbers, but I might have only captured them at one of them to avoid duplicate too many comments.

Lord-Grey (commented):
@bhill00 Many thanks for sharing a PR on supporting Wayland related screen capture capabilities. This is very much appreciated.
I put some review comments after an initial run through the PR.

I also tried to get  the desktop-portal grabber running,  but unfortunately the code fails on an Ubuntu 24.04 build.
I am asked to share my screen  which nicely creates the token.
Afterwards  the app cores.-


```
QT_LOGGING_CONF=./qtlogging.ini ./build/bin/hyperiond  -d -u ./config/wayland |grep DESKTOP-PORTAL
2026-08-24T22:40:25.983 |__| DESKTOP-PORTAL    : <DEBUG> Grabber.cpp:171:setWidthHeight() | Set new width: 1920, height: 1080 for capture
2026-08-24T22:40:25.983 |__| DESKTOP-PORTAL    : <DEBUG> Grabber.cpp:182:setFramerate() | Set new frames per second to: 1 fps, current fps: 25
2026-08-24T22:40:25.983 |__| DESKTOP-PORTAL    : <INFO> Set new frames per second to: 1 fps
2026-08-24T22:40:25.983 |__| DESKTOP-PORTAL    : <INFO> Capture interface is now enabled
2026-08-24T22:40:25.985 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:291:callPortalRequest() | [desktop-portal] CreateSession: predicted request path /org/freedesktop/portal/desktop/request/1_230/hyperion_dp_36381_1, subscribe ok=1
2026-08-24T22:40:25.987 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:309:callPortalRequest() | [desktop-portal] CreateSession: actual returned request path /org/freedesktop/portal/desktop/request/1_230/hyperion_dp_36381_1
2026-08-24T22:40:25.998 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:291:callPortalRequest() | [desktop-portal] SelectSources: predicted request path /org/freedesktop/portal/desktop/request/1_230/hyperion_dp_36381_2, subscribe ok=1

Hyperion Daemon caught signal: SIGSEGV
2026-08-24T22:40:26.007 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:309:callPortalRequest() | [desktop-portal] SelectSources: actual returned request path /org/freedesktop/portal

```

And the stack trace

```
#0  0x00006042993a5b52 in spa_pod_builder_addv(spa_pod_builder *, typedef __va_list_tag __va_list_tag *) (builder=0x701885ffa1a0, args=0x701885ff9f60) at /usr/include/spa-0.2/spa/pod/builder.h:600
600			choice = *format == '?';
[Current thread is 1 (Thread 0x701885ffb6c0 (LWP 36403))]
(gdb) bt
#0  0x00006042993a5b52 in spa_pod_builder_addv(spa_pod_builder *, typedef __va_list_tag __va_list_tag *) (builder=0x701885ffa1a0, args=0x701885ff9f60) at /usr/include/spa-0.2/spa/pod/builder.h:600
#1  0x00006042993a671e in spa_pod_builder_add (builder=0x701885ffa1a0) at /usr/include/spa-0.2/spa/pod/builder.h:627
#2  0x00006042993abf47 in DesktopPortalGrabber::runStream (this=0x6042cdecad70) at /home/thomas/development/repos/hyperion.ng-master/libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp:756
#3  0x00006042993ac3ac in DesktopPortalGrabber::portalThreadMain (this=0x6042cdecad70) at /home/thomas/development/repos/hyperion.ng-master/libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp:849
#4  0x00006042993ba010 in std::__invoke_impl<void, void (DesktopPortalGrabber::*)(), DesktopPortalGrabber*> (__f=@0x6042cdfe0e00: (void (DesktopPortalGrabber::*)(DesktopPortalGrabber * const)) 0x6042993ac2fc <DesktopPortalGrabber::portalThreadMain()>, 
    __t=@0x6042cdfe0df8: 0x6042cdecad70) at /usr/include/c++/13/bits/invoke.h:74
#5  0x00006042993b9f63 in std::__invoke<void (DesktopPortalGrabber::*)(), DesktopPortalGrabber*> (__fn=@0x6042cdfe0e00: (void (DesktopPortalGrabber::*)(DesktopPortalGrabber * const)) 0x6042993ac2fc <DesktopPortalGrabber::portalThreadMain()>)
    at /usr/include/c++/13/bits/invoke.h:96
#6  0x00006042993b9ec3 in std::thread::_Invoker<std::tuple<void (DesktopPortalGrabber::*)(), DesktopPortalGrabber*> >::_M_invoke<0ul, 1ul> (this=0x6042cdfe0df8) at /usr/include/c++/13/bits/std_thread.h:292
#7  0x00006042993b9e78 in std::thread::_Invoker<std::tuple<void (DesktopPortalGrabber::*)(), DesktopPortalGrabber*> >::operator() (this=0x6042cdfe0df8) at /usr/include/c++/13/bits/std_thread.h:299
#8  0x00006042993b9e58 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (DesktopPortalGrabber::*)(), DesktopPortalGrabber*> > >::_M_run (this=0x6042cdfe0df0) at /usr/include/c++/13/bits/std_thread.h:244
#9  0x00007018bbcecdb4 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00007018bb89cb84 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#11 0x00007018bb929d6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

bhill00:
Thanks so much for the thorough review, @Lord-Grey - really appreciate you taking the time to actually build and run this, not just read the diff. I worked through your feedback with Claude Code (I'm using it as a coding assistant on this project), reviewing and testing everything myself along the way. Finding that SIGSEGV in particular was a big help; I don't think we'd have caught it without you testing on a different distro/SPA version than mine.

Here's what we found and fixed, pushed in the latest commit:

- The crash + the SPA_POD_Propf compile error turned out to be the same root cause - that macro isn't declared in some distros' older SPA headers, and the "obvious" fix (dropping the trailing "f") silently changes the argument shape passed into spa_pod_builder_addv(), which is what crashed for you. We spelled out the exact portable expansion by hand instead, and I confirmed on my own NVIDIA setup that the desktop-portal grabber now negotiates and captures cleanly where it used to crash at that exact point.
- Switched std::mutex/condition_variable to QMutex/QWaitCondition in both grabbers to match the rest of the codebase - thanks for catching that, you're right that it was inconsistent.
- Fixed the restore-token path to use QStandardPaths::AppDataLocation, following the same pattern as ProviderRestApi::matchesPinnedCertificate() - appreciate you pointing me at that example.
- Capped the reported fps list at the actual 30fps ceiling, downgraded routine disconnects (game closed, permission revoked) to Info instead of Warning, and applied your suggested rewrite of evalScreenGrabberType() - that's a lot cleaner, thank you.
- Fixed the CMakeLists.txt indentation.

A few points we dug into and are pushing back on slightly - though happy to just do it your way if you feel strongly, you know this codebase far better than we do:

- setInError()/_isDeviceInError for the PipeWire core/stream error callbacks: both grabbers deliberately treat "no session available right now" (gamescope not running, portal permission not granted) as an expected, routine condition they just keep retrying against forever, not a device fault - that's consistent with the "_isAvailable = true unconditionally" comment already in each constructor. Flagging _isDeviceInError here felt like it would misrepresent a normal state (e.g. the user just closed their game) as a hardware/device problem. If you'd rather see it flagged anyway, tell me which specific callsites and I'll add it.
- The destructor/stop() lifecycle question: we traced the actual ownership chain - GamescopeGrabber/DesktopPortalGrabber are plain value members (not QObjects) of their Wrapper, and the Wrapper itself is held in a QScopedPointer, not deleteLater()'d. That means destruction is standard synchronous C++ RAII, so the Qt deferred-deletion concern doesn't look like it applies here. If you know of a path where the Wrapper does get torn down via deleteLater() elsewhere that we've missed, let me know and I'll add the TRACK_SCOPE() tracing to confirm it empirically.
- Requesting RGB32 directly from the capture instead of going through ImageResampler: a genuinely good perf idea, but we didn't want to swap out a path we'd just gotten working correctly across all the negotiated formats (BGRx/RGBx/BGRA) for hand-rolled per-format conversion without more careful testing than we could responsibly do in this pass. I'd rather take this on as a focused follow-up than risk a subtle color/channel bug slipping into something that currently works. If it's a blocker for you, I'll give it a shot now instead.
- Moving the Qt::DBus find_package into the grabber's own CMakeLists.txt: ENABLE_DESKTOP_PORTAL gates whether libsrc/grabber/CMakeLists.txt even add_subdirectory()s desktopportal/ at all, so the availability probe has to run before that point - fully relocating it would mean making the subdirectory inclusion unconditional and propagating the result back up via PARENT_SCOPE, which felt like a bigger structural change than this PR should carry. If that's the direction you'd want for this codebase regardless, I'll take a shot at it.

Thanks again for such a careful look at this - let me know what you think on the points above.

.github/workflows/push_pull.yml
```diff
@@ -22,17 +22,17 @@ jobs:
       fail-fast: false
       matrix:
         debian: [
-          { codename: [ 'bullseye', 'Bullseye' ] },
           { codename: [ 'bookworm', 'Bookworm' ] },
-          { codename: [ 'trixie', 'Trixie' ] }
+          { codename: [ 'trixie', 'Trixie' ] },
+          { codename: [ 'forky', 'Forky' ] }
```
Paulchen-Panther: Unfortunately, the creation of the Debian Forky Docker image for the armv6hf architecture is inconsistent. The Debian team is still making too many changes, which means the Docker container doesn't always run successfully. The only solution is to wait.

Lord-Grey: OK. I will remove the „Forky“ builds for now.

libsrc/grabber/desktopportal/DesktopPortalGrabber.cpp
```diff
@@ -0,0 +1,947 @@
...
+	spa_rectangle rectDefault = SPA_RECTANGLE(320, 240);
+	// Tried constraining this down (e.g. to 960x540) to get the compositor to scale down before
+	// handing us the buffer - KWin's screencast implementation refused entirely ("no more input
+	// formats"), meaning it only offers native resolution here. Reverted to accepting whatever
+	// native size is offered; see onStreamProcess()'s row/column-strided read below for the
+	// actual fix to the CPU cost this causes.
+	spa_rectangle rectMax = SPA_RECTANGLE(4096, 4096);
+
+	spa_fraction rateMin = SPA_FRACTION(0, 1);
+	spa_fraction rateDefault = SPA_FRACTION(25, 1);
```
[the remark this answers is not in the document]
Lord-Grey: Done

Lord-Grey:
@bhill00 I needed some more time to add tracing and do further testing to share additional findings...
I only added to the portal grabber for now to help getting that one running in a clean way. We can apply tracing and corrections then to the gamescope one in a 2nd step.

To trace specific screen grabber aspects, the best way is to enable the to be traced elements in the qtlogging.ini you find in the repository's root directory.
Depending how you run hyperiond you can enable the configuration via `QT_LOGGING_CONF=./qtlogging.ini`

or if you execute from the cmd-line you can do
`QT_LOGGING_CONF=./qtlogging.ini ./build/bin/hyperiond  -d -u ./config/pipe`

Here is a sample to activate the elements traceable for screen grabbers:
```
# Grabber logging
#hyperion.grabber.flow.debug=true
#hyperion.grabber.screen.*.debug=true
hyperion.grabber.screen.capture.debug=true
hyperion.grabber.screen.flow.debug=true
hyperion.grabber.screen.properties.debug=true
hyperion.grabber.screen.benchmark.debug=true
```

It might be also helpful to track the image object too, to get an understanding of the inetractione between grabber and image:
```
# Object Image/ImageData logging
#hyperion.image.*.debug=true
hyperion.image.create.debug=true
#hyperion.image.copy.debug=true
#hyperion.image.move.debug=true
#hyperion.image.assign.debug=true
#hyperion.image.destroy.debug=true
hyperion.image.track.debug=true
```

Object tracing is helpful esp. to see objects constructed or destroyed. It follows the "TRACK_SCOPE()":

```
# Object tracing
hyperion.objects.track.debug=true
```

In summary, uncomment the aspect you would like to trace....

Lord-Grey:
Now with tracing on, following findings:

Scenario: Grabber is disabled (in the UI)

image
 
1a. While the grabber is disabled, it starts and grabs 
Expectation: the grabber should not waste resources

```
2026-09-19T16:31:39.673 |__| DESKTOP-PORTAL    : <INFO> Capture interface is now disabled
2026-09-19T16:31:39.673 |__| DAEMON            : <INFO> DesktopPortal grabber created
...
2026-09-19T16:31:39.674 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] CreateSession: predicted request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_1, subscribe ok=1
2026-09-19T16:31:39.676 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] CreateSession: actual returned request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_1
..
2026-09-19T16:31:39.732 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::readRestoreToken()[311] TID:17836 (hyperiond) Reading restore token from ".../token/desktop-portal-restore-token"
...
2026-09-19T16:31:39.733 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] SelectSources: predicted request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_2, subscribe ok=1
...
2026-09-19T16:31:39.747 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] SelectSources: actual returned request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_2
2026-09-19T16:31:39.747 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] Start: predicted request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_3, subscribe ok=1
2026-09-19T16:31:39.748 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] Start: actual returned request path /org/freedesktop/portal/desktop/request/1_146/hyperion_dp_17813_3
2026-09-19T16:31:39.811 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::writeRestoreToken()[333] TID:17836 (hyperiond) Writing restore token to "/home/thomas/.local/share/Hyperion Daemon/token/desktop-portal-restore-token"
2026-09-19T16:31:39.812 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::connectCore()[765] TID:17836 (hyperiond) Connecting to Pipewire core
2026-09-19T16:31:39.813 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::runStream()[795] TID:17836 (hyperiond) Running Pipewire stream
2026-09-19T16:31:39.813 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[500] TID:17836 (hyperiond) Stream state changed to "CONNECTING" , error: (none)
2026-09-19T16:31:39.813 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[500] TID:17836 (hyperiond) Stream state changed to "PAUSED" , error: (none)
2026-09-19T16:31:39.815 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[540] TID:17836 (hyperiond) Stream param changed, id= 2 , param= valid
2026-09-19T16:31:39.815 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[540] TID:17836 (hyperiond) Stream param changed, id= 15 , param= valid
2026-09-19T16:31:39.815 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[540] TID:17836 (hyperiond) Stream param changed, id= 4 , param= valid
2026-09-19T16:31:39.815 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[575] TID:17836 (hyperiond) Stream format changed to 3840 x 2160 , modifier: 0 , format: 8
2026-09-19T16:31:39.965 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[500] TID:17836 (hyperiond) Stream state changed to "STREAMING" , error: (none)
2026-09-19T16:31:39.992 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[637] TID:17836 (hyperiond) "Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
2026-09-19T16:31:39.992 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[669] TID:17836 (hyperiond) "Buffer data type=2 (MemFd=2 DmaBuf=3 MemPtr=1)"
2026-09-19T16:31:40.035 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[637] TID:17836 (hyperiond) "Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
```

1b. If no token is available yet, the user is asked for approving screen capture.
Expectation: This must not happen, when the grabber is stopped. 

1c. If I cancel the "screen share" request, I am asked again and again for approval, even if the grabber is stopped
Expectation: This must not happen, when the grabber is stopped.

```
2026-09-19T16:55:57.494 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::negotiatePortalSession()[392] TID:19763 (hyperiond) Negotiating portal session
2026-09-19T16:55:57.495 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] CreateSession: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_5, subscribe ok=1
2026-09-19T16:55:57.498 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] CreateSession: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_5
2026-09-19T16:55:57.498 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] SelectSources: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_6, subscribe ok=1
2026-09-19T16:55:57.499 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] SelectSources: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_6
2026-09-19T16:55:57.500 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] Start: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_7, subscribe ok=1
2026-09-19T16:55:57.500 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] Start: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_7
2026-09-19T16:56:00.041 |__| DESKTOP-PORTAL    : <WARNING> [desktop-portal] Start failed or was denied (response code 1)
2026-09-19T16:56:00.041 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::closePortalSession()[484] TID:19763 (hyperiond) Closing portal session
2026-09-19T16:56:00.041 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::waitBeforeRetry()[935] TID:19763 (hyperiond) Waiting before retrying portal session
2026-09-19T16:56:02.041 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::negotiatePortalSession()[392] TID:19763 (hyperiond) Negotiating portal session
2026-09-19T16:56:02.042 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] CreateSession: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_9, subscribe ok=1
2026-09-19T16:56:02.044 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] CreateSession: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_9
2026-09-19T16:56:02.045 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] SelectSources: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_10, subscribe ok=1
2026-09-19T16:56:02.046 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] SelectSources: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_10
2026-09-19T16:56:02.046 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:359:callPortalRequest() | [desktop-portal] Start: predicted request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_11, subscribe ok=1
2026-09-19T16:56:02.047 |__| DESKTOP-PORTAL    : <DEBUG> DesktopPortalGrabber.cpp:377:callPortalRequest() | [desktop-portal] Start: actual returned request path /org/freedesktop/portal/desktop/request/1_162/hyperion_dp_19742_11
2026-09-19T16:56:07.033 |__| DESKTOP-PORTAL    : <WARNING> [desktop-portal] Start failed or was denied (response code 1)

```

Lord-Grey:
2. Grabber is enabled and capture frequency = 1s to be able to follow the activities... Size decimation = 8

image

2a. During processing, you see that the various images related sizes are different and are not aligned, which result in multiple scaling processing activities.

Screen size: Width: 3840, Height: 2160
In onStreamProcess the full screen size image is captured as 
"Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
and then sampled by a hard factor down by 4 (why?)
Sampled resolution: 960 x 540

Then  "Frontbuffer; Width: 960, Height: 540, PixelFormat: BGR32" will be scaled into the final output image:
Frame grabbed into image with size: 120 x 67
which means that the final image is de facto downscaled by factor 32 and not 8 as given by size decimation.
With size decimation = 4, the image is still 4 times smaller then the captured screen.

Why is the screen not already captured in full screen size divided by size decimation?
To best achieve this would during discovery getting the real screen size and not an artificial 1920x1080 (with the assumption that the screensize is not likely to change. If it does  onstreamchange can handle it.
Then you would have the target width already available as the grabber width, plus the output image corresponds in size.


```
2026-09-19T18:22:33.232 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::connectCore()[769] TID:25982 (hyperiond) Connecting to Pipewire core
2026-09-19T18:22:33.233 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::runStream()[799] TID:25982 (hyperiond) Running Pipewire stream
2026-09-19T18:22:33.233 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[502] TID:25982 (hyperiond) Stream state changed to "CONNECTING" , error: (none)
2026-09-19T18:22:33.234 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[502] TID:25982 (hyperiond) Stream state changed to "PAUSED" , error: (none)
2026-09-19T18:22:33.236 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[542] TID:25982 (hyperiond) Stream param changed, id= 2 , param= valid
2026-09-19T18:22:33.236 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[542] TID:25982 (hyperiond) Stream param changed, id= 15 , param= valid
2026-09-19T18:22:33.236 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[542] TID:25982 (hyperiond) Stream param changed, id= 4 , param= valid
2026-09-19T18:22:33.236 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamParamChanged()[577] TID:25982 (hyperiond) Stream format changed to 3840 x 2160 , modifier: 0 , format: 8
2026-09-19T18:22:33.354 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamStateChanged()[502] TID:25982 (hyperiond) Stream state changed to "STREAMING" , error: (none)
...
2026-09-19T17:46:11.089 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[637] TID:21871 (hyperiond) "Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
2026-09-19T17:46:11.646 |--|                   : <TRACE> hyperion.image.track GrabberWrapper::transferFrame()[118] TID:21852 (hyperiond) Image [ 1 ], resizing image from  120 x 67  to  960 x 540
2026-09-19T17:46:11.647 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::grabFrame()[231] TID:21852 (hyperiond) "Frontbuffer; Width: 960, Height: 540, PixelFormat: BGR32"
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track CaptureCont::handleScreenImage()[112] TID:21864 (HyperionThread) Image [ 520 ], name: DesktopPortal  image size: 120 x 67
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track Hyperion::setInputImage()[448] TID:21864 (HyperionThread) |I0| setInputImage()... Image [ 520 ], priority 250 image id 520 timeout -1 ms
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track PriorityMuxer::setInputImage()[297] TID:21864 (HyperionThread) Image [ 520 ], Priority: 250 ,timeout: -1 ms
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track PriorityMuxer::setInputImage()[333] TID:21864 (HyperionThread) Image [ 520 ] assigned to priority: 250 ,timeout: -1 ms
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track ImageProcessor::process()[129] TID:21864 (HyperionThread) Image [ 522 ]
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track hyperion::BlackBorderProcessor::process()[69] TID:21864 (HyperionThread) Image [ 522 ]
2026-09-19T17:46:11.648 |--|                   : <TRACE> hyperion.image.track hyperion::BlackBorderDetector::process()[69] TID:21864 (HyperionThread) Image [ 522 ]
```

Lord-Grey:
3. Grabber is enabled as in 2. 

3a. Disable Grabber
Expectation: Grabbing stops
Actual: onStreamProcess continues to capture screens.

```
2026-09-19T18:32:53.982 |--|                   : <TRACE> hyperion.grabber.screen.flow GrabberWrapper::handleSourceRequestScreen()[354] TID:26729 (hyperiond) Instance [ 0 ] - Request to remove screen grabber "DesktopPortal" which is enabled
2026-09-19T18:32:53.982 |--|                   : <TRACE> hyperion.grabber.screen.flow GrabberWrapper::handleSourceRequestScreen()[378] TID:26729 (hyperiond) Instance [ 0 ] - Removing screen grabber "DesktopPortal"
2026-09-19T18:32:53.982 |__| GRABBER-DESKTOPPO : <DEBUG> GrabberWrapper.cpp:383:handleSourceRequestScreen() | Stop screen grabber DesktopPortal, as no instance is listing any longer
2026-09-19T18:32:53.982 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::stop()[115] TID:26729 (hyperiond) Request to stop grabber "DesktopPortal" , which is currently enabled , and active
2026-09-19T18:32:53.982 |__| GRABBER-DESKTOPPO : <INFO> DesktopPortal grabber stopped
2026-09-19T18:32:53.982 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::stop()[125] TID:26729 (hyperiond) Grabber "DesktopPortal" stopped
..
2026-09-19T18:32:53.997 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::handleSettingsUpdate()[298] TID:26729 (hyperiond) Screen grabber "DesktopPortal" updating settings with QJsonObject({"available_devices":"Desktop Portal","cropBottom":0,"cropLeft":0,"cropRight":0,"cropTop":0,"device":"desktop-portal","device_inputs":"0","enable":false,"fps":1,"framerates":"1","height":1080,"input":0,"pixelDecimation":8,"resolutions":"0","width":1920})
2026-09-19T18:32:53.998 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::handleSettingsUpdate()[305] TID:26729 (hyperiond) Screen grabber "DesktopPortal" is configured disabled , current state is enabled
2026-09-19T18:32:53.998 |__| DESKTOP-PORTAL    : <INFO> Capture interface is now disabled
2026-09-19T18:32:53.998 |__| GRABBER-DESKTOPPO : <INFO> Stop running grabber DesktopPortal after settings update
2026-09-19T18:32:53.998 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::stop()[115] TID:26729 (hyperiond) Request to stop grabber "DesktopPortal" , which is currently disabled , and inactive
2026-09-19T18:32:53.998 |--|                   : <TRACE> hyperion.grabber.flow GrabberWrapper::stop()[125] TID:26729 (hyperiond) Grabber "DesktopPortal" stopped
2026-09-19T18:32:54.015 |I0| MUXER             : <DEBUG> PriorityMuxer.cpp:406:updatePriorities() | Removed source priority 250
2026-09-19T18:32:54.015 |I0| MUXER             : <DEBUG> PriorityMuxer.cpp:469:updatePriorities() | Set visible priority to 255
2026-09-19T18:32:54.015 |I0| HYPERION          : <DEBUG> Hyperion.cpp:681:handleSourceAvailability() | No source left -> Pause output processing and switch LED-Device off
2026-09-19T18:32:54.015 |I0| LEDDEVICE         : <INFO> Switching device file OFF
2026-09-19T18:32:54.015 |I0| LEDDEVICE         : <INFO> Device file is OFF
2026-09-19T18:32:54.898 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[639] TID:26750 (hyperiond) "Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
2026-09-19T18:32:54.898 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamProcess()[733] TID:26750 (hyperiond) Sampled resolution: 960 x 540
2026-09-19T18:32:56.104 |--|                   : <TRACE> hyperion.grabber.screen.capture DesktopPortalGrabber::onStreamProcess()[639] TID:26750 (hyperiond) "Processing frame, Width: 3840, Height: 2160, SpaFormat: 8, PixelFormat: BGR32"
2026-09-19T18:32:56.104 |--|                   : <TRACE> hyperion.grabber.screen.flow DesktopPortalGrabber::onStreamProcess()[733] TID:26750 (hyperiond) Sampled resolution: 960 x 540
```

